PR #4, opened 2017-03-21
Shapeshift.iot Integration

iousowp:
Integration of Shapeshift.io for accepts altcoins in Woocommerce

### DIFF
--- a/order.php
+++ b/order.php
@@ -94,6 +94,16 @@ get_header();
           Transaction : <a style="font-weight:bold;color: #956431;" target="_blank"
                            href="http://www.blockonomics.co/api/tx?txid={{order.txid}}&addr={{order.address}}">{{order.txid |limitTo: 20}}</a>
         </div>
+        <div>
+<h5>Pay with altcoins</h5>
+   <?php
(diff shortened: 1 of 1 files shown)


blockonomics:
Thanks. I will review this and get back in a day.

Can you  please do the following:

- Make this an option in settings , enable Pay with altcoins 
- Default value of option will be false
- If option is enabled, pay with altcoin will be shown, else not shown

astupidmoose:
FYI, This integration defines the $process variable, which is actually the authors API key with shapeshift. Therefore the author of this integration is automatically getting paid affiliate referrals for integrations. 

This should probably be commented somewhere to let people who use it know.

blockonomics:
@astupidmoose Thanks for pointing this out. The pull request has ofcourse not been merged. If you have time can you create a pull request without the affiliate and I would be happy to merge it in.

astupidmoose:
Ideally this would be an optional field. As stated above, users should have the ability to turn on/off the integration. Once turned on, user should be able to add their own public key for shapeshift affiliate credit. 

Unfortunately I do not know enough about wordpress plugins to do this myself, however I thought I would post this warning.

PR #24, opened 2017-10-18
Update translation files

aleangel1212:


aleangel1212:
Updated Spanish translation to new strings and added French and German translations

PR #56, opened 2018-03-02
Intelligent error display on plugin

jusasiiv:
Verbose error handling

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -279,7 +279,12 @@ if (is_plugin_active('woocommerce/woocommerce.php') || class_exists('WooCommerce
 
             private function checkForErrors($responseObj, $woocommerce) {
                 
-                if(!ini_get('allow_url_fopen')) {
+                if(!isset($responseObj->response_code)) {
+                    $error_str = __('Your webhost is blocking outgoing HTTPS connections. Blockonomics requires an outgoing HTTPS POST (port 443) to generate new address. Check with your webhosting provider to allow this.', 'blockonomics-bitcoin-payments');
+                    $this->displayError($error_str, $woocommerce);
+                    return false;
+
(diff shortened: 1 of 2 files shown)


blockonomics-woocommerce.php
```diff
@@ -156,29 +156,62 @@ public function process_payment($order_id)
...
-                try {
-                    $blockonomics = new Blockonomics;
-                    $address = $blockonomics->new_address(get_option('blockonomics_api_key'), get_option("blockonomics_callback_secret"));
-                    $price = $blockonomics->get_price(get_woocommerce_currency());
-                } catch (Exception $e) {
-                    $address = '';
+                $blockonomics = new Blockonomics;
+                $responseObj = $blockonomics->new_address(get_option('blockonomics_api_key'), get_option("blockonomics_callback_secret"));
+                $price = $blockonomics->get_price(get_woocommerce_currency());
+
```
shivaenigma: Would be good if the following error code is refactored into a new method, for cleaner code

shivaenigma:
All string must be put into translation construct so that these can be [translated](https://translate.wordpress.org/projects/wp-plugins/blockonomics-bitcoin-payments) into other languages, example
`_('Sorry, but there was an error processing your order. Please try again or try a different payment method. (plugin not configured)', 'blockonomics-bitcoin-payments')`

blockonomics:
Tested.  Look Ok. Please update condition to detect in case of non response obj
`
403.. API key, 503 ...blocking connection... other errors print http status code`

jusasiiv:
The response code parsing added some complexity, but if that is fine it is easy now to add conditions on internal server errors / http response errors. 

`$responseObj->{'response_code'} = $http_response_header[0];` was added because if the response was empty, the whole object was empty and could not be parsed for message

Issue #69, opened 2018-09-11
Way to auto switch to curl


blockonomics:
-We should auto switch to curl if fopen error is coming
-An idea is to try fopen/curl during test setup and then test a flag that tells us what to use.
-If nothing works, just keep showing fopen error

blockonomics:
I think we can use (wp_remote_post)(https://codex.wordpress.org/Function_Reference/) and wp_remote_get. It should automatically see that is available on the system (Cross Check)

PR #80, opened 2018-10-10
Fix for issue #69, use wp_remote for requests

DarrenWestwood:
Fix for issue  [#69](https://github.com/blockonomics/woocommerce-plugin/issues/69), [use wp_remote for requests](https://github.com/blockonomics/woocommerce-plugin/commit/ce1a65927ace89cd4b2d227adc57be63b1b99a8f)


blockonomics:
Code has becoming a lot and repetitive. Do refactoring:

check_callbacks() and be removed and put into one since testsetup() method. See [this code](https://github.com/blockonomics/prestashop-plugin/blob/master/blockonomics.php#L252)

PR #86, opened 2018-10-24
Version 1.6.2

DarrenWestwood:


blockonomics-woocommerce.php
```diff
@@ -165,6 +166,7 @@ public function process_payment($order_id)
                 }else{
                 	$price = 1;
                 }
+                $price = $price * ((100-get_option('blockonomics_margin', 0)) / 100);
```
shivaenigma: `$price = $price * 100/(100+get_option('blockonomics_margin', 0))`

shivaenigma: Also don't apply when currency is BTC . Doesn't make sense there

blockonomics-woocommerce.php
```diff
@@ -573,11 +575,15 @@ function checkForAPIKeyChange()
                         </select>
                     </td>
                 </tr>
+                <tr valign="top">
+                    <th scope="row"><?php echo __('Currency Rate Margin % (Increase fiat to BTC calculated rate by this percentage)', 'blockonomics-bitcoin-payments')?></th>
+                    <td><input type="number" min="0" max="5" name="blockonomics_margin" value="<?php echo esc_attr( get_option('blockonomics_margin', 0) ); ?>" /></td>
```
shivaenigma: Lets put max as 4

PR #104, opened 2019-02-07
One click setup

jusasiiv:
One click setup changes for review

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -151,32 +151,9 @@ if (is_plugin_active('woocommerce/woocommerce.php') || class_exists('WooCommerce
                     $message = __('Congrats ! Setup is all done', 'blockonomics-bitcoin-payments');
                     display_admin_message($message, 'updated');
 
-                    // Only make withdraw if temp api key is not null
-                    if(get_option('blockonomics_temp_api_key'))
-                    {
-                        // Check if merchant has received any confirmed BTC payments
-                        // If there are funds, make a withdraw request. Only set temp api key to null if success
-                        if (get_option('temp_withdraw_amount') > 0)
-                        {
-                            $response = $blockonomics->make_withdraw();
-                            if ($response->response_code != 200)
-                            {
-                                $message = __('Error while making withdraw: '. $response->message, 'blockonomics-bitcoin-payments');
-                                display_admin_message($message, 'error');
-                            }
-                            else
-                            {
-                                $message = __('Your funds withdraw request has been submitted. Please check your Blockonomics registered emailid for details', 'blockonomics-bitcoin-payments');
-                                display_admin_message($message, 'updated');
-                                update_option("blockonomics_temp_api_key", null);
-                                update_option('temp_withdraw_amount', 0);
-                            }
-                        }
-                        // No funds in account, no need to withdraw
-                        else
-                        {
-                            update_option("blockonomics_temp_api_key", null);
-                        }
+                    $message = $blockonomics->make_withdraw();
+                    if ($message) {
+                        display_admin_message($message[0], $message[1]);
                     }
                 }
             }
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -104,15 +104,14 @@ if (is_plugin_active('woocommerce/woocommerce.php') || class_exists('WooCommerce
             {
                 generate_secret(true);
             }
-            
+
             $api_key = $blockonomics->get_api_key();
             
             // get_api_key() will return api key or temp api key
             // if both are null, generate new blockonomics guest account with temporary wallet
             // temp wallet will be used with temp api key
             if (!$api_key)
-            { 
-                display_admin_message('Null api key found', 'error');
+            {
                 generate_secret();
                 $callback_url = get_callback_url();
                 $response = $blockonomics->get_temp_api_key($callback_url);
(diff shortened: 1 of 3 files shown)


blockonomics-woocommerce.php
```diff
@@ -115,93 +141,116 @@ function add_page()
...
+        {
+            $callback_secret = get_option('blockonomics_callback_secret');
+            $callback_url = WC()->api_request_url('WC_Gateway_Blockonomics');
+            $callback_url = add_query_arg('secret', $callback_secret, $callback_url);
+            return $callback_url;
+        }
+
+        function get_total_received()
+        {
+            $blockonomics_orders = get_option('blockonomics_orders');
```
jusasiiv: What if there are under or over payments? Using this we only see the requested amounts

blockonomics-woocommerce.php
```diff
@@ -98,9 +98,35 @@ function woocommerce_handle_blockonomics_return()
...
+            // temp wallet will be used with temp api key
+            if ($api_key == null)
+            {
+                generate_secret();
+                $callback_url = get_callback_url();
+                $response = $blockonomics->get_temp_api_key($callback_url);
+
+                if ($response->response_code != 200)
+                {
+                    $message = __('Error while generating temporary APIKey: '. $response->message, 'blockonomics-bitcoin-payments');
```
jusasiiv: What if error does not have message? For example 401 or other

blockonomics:
1. Temp API key creation should be in activation function

blockonomics-woocommerce.php
```diff
@@ -98,9 +98,35 @@ function woocommerce_handle_blockonomics_return()
...
+            }
+
+            $api_key = $blockonomics->get_api_key();
             
-            generate_secret();
-            register_setting('blockonomics_g', 'blockonomics_gen_callback', 'gen_callback');
+            // get_api_key() will return api key or temp api key
+            // if both are null, generate new blockonomics guest account with temporary wallet
+            // temp wallet will be used with temp api key
+            if ($api_key == null)
```
blockonomics: I don't like comparison with `null` .. can we use if (!$api_key).  It should cover empty API key case also

blockonomics-woocommerce.php
```diff
@@ -109,99 +135,133 @@ function add_page()
...
                 }
                 else
                 {
                     $message = __('Congrats ! Setup is all done', 'blockonomics-bitcoin-payments');
-                    $type = 'updated';
-                    add_settings_error('option_notice', 'option_notice', $message, $type);
+                    display_admin_message($message, 'updated');
+
+                    // Only make withdraw if temp api key is not null
+                    if(get_option('blockonomics_temp_api_key') != null)
```
blockonomics: Lets avoid null comparisons

php/Blockonomics.php
```diff
@@ -42,36 +54,50 @@ public function new_address($api_key, $secret, $reset=false)
...
+        $body = json_encode(array('callback' => $callback_url));
+        $response = $this->post($url, '', $body);
+        $responseObj = json_decode(wp_remote_retrieve_body($response));
+        $responseObj->{'response_code'} = wp_remote_retrieve_response_code($response);
+        return $responseObj;
+    }
+
+    public function make_withdraw()
+    {
+        $api_key = get_option("blockonomics_api_key");
```
blockonomics: use api_key from object instance

php/Blockonomics.php
```diff
@@ -42,36 +54,50 @@ public function new_address($api_key, $secret, $reset=false)
...
+        $temp_api_key = get_option('blockonomics_temp_api_key');
```
blockonomics: Do some sanity checks.. if temp_api_key is empty or api_key doesn't exist ... then do nothing

php/WC_Gateway_Blockonomics.php
```diff
@@ -208,6 +206,12 @@ public function check_blockonomics_callback()
                         $wc_order->add_order_note(__('Payment completed', 'blockonomics-bitcoin-payments'));
                         $wc_order->payment_complete($order['txid']);
                     }
+                    // Keep track of funds in temp wallet
+                    if(get_option('blockonomics_temp_api_key')) {
+                        $current_temp_amount = get_option('temp_withdraw_amount');
```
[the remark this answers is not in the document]
jusasiiv: `get_option()` returns False if there is no value. PHP handles False as 0 if making calculations so here we can assume no option = 0

blockonomics-woocommerce.php
blockonomics: if get_temp_api_key doesn't work. Its probably a critical error. Very less chance of happening

php/Blockonomics.php
```diff
@@ -42,36 +54,66 @@ public function new_address($api_key, $secret, $reset=false)
...
+        $responseObj = json_decode(wp_remote_retrieve_body($response));
+        $responseObj->{'response_code'} = wp_remote_retrieve_response_code($response);
+        return $responseObj;
+    }
+
+    public function make_withdraw()
+    {
+        $api_key = $this->api_key;
+        $temp_api_key = get_option('blockonomics_temp_api_key');
+        if (!$api_key || !$temp_api_key) {
```
blockonomics: Also check if `$temp_api_key != $api_key`

php/Blockonomics.php
```diff
@@ -8,16 +8,28 @@ class Blockonomics
...
 
     public function __construct()
     {
+        $this->api_key = $this->get_api_key();
+    }
+
+    public function get_api_key()
+    {
+        $api_key = get_option("blockonomics_api_key");
+        if ($api_key == null)
```
blockonomics: Lets avoid null comparison

PR #111, opened 2019-02-13
Altcoin refunds

DarrenWestwood:
Added spinner
Fix low/high error

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -334,38 +334,38 @@ app.controller('AltcoinController', function($scope, $interval, Order, AltcoinNe
                 //Wait for both the altcoin limits and new altcoin order uuid
                 Promise.all(promises)
                     .then(values => {
-                        var alt_minimum = values[0].min;
-                        var alt_maximum = values[0].max;
-                        //Compare the min/max limits for altcoin payments with the order amount
-                        if(amount <= alt_minimum) {
-                            //Order amount too low for altcoin payment
-                            update_altcoin_status('low_high');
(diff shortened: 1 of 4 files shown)


templates/track.php
```diff
@@ -10,6 +10,8 @@
           </div>
         </div>
       </div>
+      <!-- Spinner -->
+      <div class="bnomics-spinner" ng-show="spinner"><div class="bnomics-ring"><div></div><div></div><div></div><div></div></div></div>
```
shivaenigma: add ng-cloak attribute

js/app.js
```diff
@@ -334,38 +336,38 @@ app.controller('AltcoinController', function($scope, $interval, Order, AltcoinNe
...
-                            //Accept the altcoin order using the uuid
-                            AltcoinAccept.save({
-                                    "uuid": uuid
-                                },function(order_accept) {
-                                    //Fetch the order details using bitcoin address
-                                    Order.get({
-                                        "get_order": order_accept.order.destination
-                                    }, function(order) {
+                        //Fetch the order details using bitcoin address
+                        Order.get({
```
shivaenigma: This call is uncessary. Only thing missing is orderid, Lets get also as parameter of create_order

shivaenigma:
- Added two comments
- Also in `info_order(uuid) method` . Extra `fetch_order_id` is not required. Get_order is already returning orderid. We can use `fetch_order_id` method from server as well if its not used elsewhere

Issue #124, opened 2019-04-19
checkout intermittently not working


blockonomics:
- Gives unable to generate address
- I experienced this sometimes and one customer complained.

This looks like a timeout issue when doing new_address call. Check timeout parameter for wp_post and increase the value

blockonomics:
- Lets reproduce this issue first by decreasing the timeout of _wp_remote_post_
- If able to reproduce then to fix... we can increase timeout to 10s

DarrenWestwood:
The error is reproducible with the timeout set to 1s

blockonomics:
- Ok looks like server is sometime busy and may not respond. 
- Default timeout is 5sec (should be enough), still lets increase to new_address call timeout to 8sec

PR #132, opened 2019-05-28
Wp remote post timeout

DarrenWestwood:
Issue #124, increase wp_remote_post timeout to 8s for new address.


php/Blockonomics.php
```diff
@@ -129,12 +129,13 @@ private function get($url, $api_key = '')
         }
     }
 
-    private function post($url, $api_key = '', $body = '', $type = '')
```
blockonomics: Let us not expicitly use the default timeout of 5 sec, since later wp_remote_post could change default. 
When using default don't set the parameter or set to null

Issue #71, opened 2018-09-14
Minification/Size Issues


blockonomics:
- Currently doesn't work along with  minification plugin
- Lots of http requests are sent to load files and is taking time

Try to make it work with minifiers or combine all JS into one file

shivaenigma:
- Increasing priority of this. Try to decrease number of files/size

shivaenigma:
- Also remove unused files. It is unnecessary increasing plugin size

shivaenigma:
- Also remove unused files. It is uncessary increasing plugin size

blockonomics:
People are reporting it is taking 15 seconds to load. We need to check stats on a medium speed connection

DarrenWestwood:
[JavaScript Speed Report - Blockonomics WooCommerce.docx](https://github.com/blockonomics/woocommerce-plugin/files/3152140/JavaScript.Speed.Report.-.Blockonomics.WooCommerce.docx)

Areas to improve:
* reduce load times - minify angular-qrcode.js
* reduce non-blocking resource loads - async param when using enqueue scripts

DarrenWestwood:
Non-blocking resource load times have been reduced using defer instead of async to maintain working functionality.

PR #133, opened 2019-05-28
Wp enqueue scripts async

DarrenWestwood:
Add defer param to enqueue scripts #71


blockonomics:
@jusasiiv please review

Issue #137, opened 2019-06-28
Lite Mode Automation


blockonomics:
## Apply Lite automatically
- Detect theme during test setup / first time plugin install and set Lite Mode if it is any of our known conflicting themes.  We should not overwrite any existing lite_mode value set by user(true/false)
- Problem: We could be too conservative for some themes or miss out on some themes

## Detect payment page problems
- When payment page doesn't render, we somehow detect this and display error on Payment Page - _Unable to render correctly, Note to Administrator: Please enable lite mode in Blockonomics plugin_

blockonomics:
Added two solutions. We may implement one of them or both depending on feasibility. Please check what is doable

DarrenWestwood:
Both solutions are valid as can be seen in https://github.com/blockonomics/woocommerce-plugin/pull/138

PR #138, opened 2019-07-03
Lite Mode Automation #137

DarrenWestwood:
Solutions for lite mode automations, #137
- 8e1b4b8 If the lite mode setting has not been set before then check if a known problem theme is being used
- a3ff8e0 Detects if app.js is executing on order.php correctly

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -298,16 +298,7 @@ function blockonomics_woocommerce_init()
                         </tr>
                         <tr valign="top">
                             <th scope="row"><?php echo __('Display Payment Page in Lite Mode (Enable this if you are having problems in rendering checkout page)', 'blockonomics-bitcoin-payments')?></th>
-                            <td><input type="checkbox" name="blockonomics_lite" value="1" <?php 
-                            $current_theme = wp_get_theme();
-                            $themes = array('Divi');
-                            foreach ($themes as $theme) {
-                              if ( $theme == $current_theme->name || $theme == $current_theme->parent_theme ) {
-                                $simple_version = true;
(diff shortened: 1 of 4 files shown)


blockonomics:
Let us remove the automatic detection in settings. May lead to false positives.

@jusasiiv  please verify that checkout page, the error is shown properly on various themes.. we can track in an excel sheet

blockonomics:
>If app.js does not load it will show error, if app.js does not load (incompatible theme crashes js) the error will show.

Good idea ... @DarrenWestwood  can we try this

DarrenWestwood:
> Let us remove the automatic detection in settings. May lead to false positives.

6be20b1

> If app.js does not load it will show error, if app.js does not load (incompatible theme crashes js) the error will show.

8f01922

jusasiiv:
On slow connection error is shown a duration before JS loads. To fix this UX issue add a CSS transition within few seconds so that error message is first not shown but transition changes opacity/display value to shown.

blockonomics:
> On slow connection error is shown a duration before JS loads. To fix this UX issue add a CSS transition within few seconds so that error message is first not shown but transition changes opacity/display value to shown.

Video of this showing intermittent screen https://drive.google.com/file/d/1HZaOlFM0lbAC6jiu-tUzaV_SF_evN0j8/view?usp=sharing

blockonomics:
working as expected. I approve the functionality .. @jusasiiv  any code changes you still want to review ?

jusasiiv:
> @jusasiiv any code changes you still want to review ?

Animation code is clean and modern way to do it. No changes required. Other code is already reviewed as good.

PR #141, opened 2019-07-30
Fix altcoin refund link

DarrenWestwood:
Fixes the link in the altcoin refund email

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -237,7 +237,7 @@ app.controller('AltcoinController', function($scope, $interval, Order, AltcoinNe
         else
             params = {};
         params.uuid = uuid;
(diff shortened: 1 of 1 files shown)


js/app.js
```diff
@@ -237,7 +237,7 @@ app.controller('AltcoinController', function($scope, $interval, Order, AltcoinNe
         else
             params = {};
         params.uuid = uuid;
-        url = window.location.pathname;
+        url = window.location.protocol + "//" + window.location.host + window.location.pathname;
```
shivaenigma: why not use window.location.href ?

DarrenWestwood: We do not require the parameters from the current URL

Issue #143, opened 2019-08-05
Refreshing altcoin order URL is creating new order


blockonomics:
- Customer will use it to refresh order status and thus creates problem
- URL should only have flyp uuid and refresh page  should work

PR #146, opened 2019-08-08
Fix refreshing altcoin order issue #143

DarrenWestwood:
Move create order to checkout controller #143


blockonomics:
You have changed position of `create_order` function. Due to this unable to see what changed in the function. Can you change code so that diff is viewable

Issue #144, opened 2019-08-05
Handle flyp order expiry


blockonomics:
On expiry. Show the following message
- If you already paid, please enter your refund address to get refund.
- On adding refund address, user will also get email through which he can track the refund status

 Add a mockup screen of UI before doing implementation

DarrenWestwood:
Expired page:
![New add refund message in expiry](https://user-images.githubusercontent.com/39666372/62601608-0a8b2b80-b8f2-11e9-8115-40834a4fc36c.png)

Add refund page:
![Add refund message from expiry](https://user-images.githubusercontent.com/39666372/62601622-1119a300-b8f2-11e9-925b-be26adcbb505.png)

blockonomics:
- If you already paid, **click here** to get a refund. Same font/appearance, so this message is more prominent and not missed by the user

DarrenWestwood:
![Order expiry](https://user-images.githubusercontent.com/39666372/62680178-e9404300-b9b6-11e9-8bc6-c9c83296a0b6.png)

![add refund](https://user-images.githubusercontent.com/39666372/62680213-feb56d00-b9b6-11e9-9098-b2f20b10fc9d.png)

PR #147, opened 2019-08-08
Flypme handle expired #144

DarrenWestwood:
Update handling of expired flyp orders to add refund address #144


blockonomics:
To test look like it also required changes from the other pull request. Will wait for that one to resolve before testing this

PR #155, opened 2019-10-10
Network Confirmation field for merchants

freefiona85:
Create a network confirmation field so Merchant can decide if they need 0,1, or 2 confirmations before the transaction goes through.

### DIFF
--- a/php/WC_Gateway_Blockonomics.php
+++ b/php/WC_Gateway_Blockonomics.php
@@ -182,7 +182,7 @@ class WC_Gateway_Blockonomics extends WC_Payment_Gateway
 
         $callback_secret = get_option("blockonomics_callback_secret");
         $secret = isset($_REQUEST['secret']) ? $_REQUEST['secret'] : "";
-		$network_confirmations=get_option("blockonomics_network_confirmation");
+		$network_confirmations=get_option("blockonomics_network_confirmation",2);
         if ($callback_secret  && $callback_secret == $secret) {
             $addr = $_REQUEST['addr'];
             $order = $orders[$addr];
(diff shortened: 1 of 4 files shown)


blockonomics:
@DarrenWestwood  please review

php/WC_Gateway_Blockonomics.php
```diff
@@ -182,6 +182,7 @@ public function check_blockonomics_callback()
 
         $callback_secret = get_option("blockonomics_callback_secret");
         $secret = isset($_REQUEST['secret']) ? $_REQUEST['secret'] : "";
+		$network_confirmations=get_option("blockonomics_network_confirmation");
```
DarrenWestwood: We can use a default value of 2. This is for upgrading users who have not clicked save after updating the plugin.

php/WC_Gateway_Blockonomics.php
```diff
@@ -195,7 +196,7 @@ public function check_blockonomics_callback()
                     $minutes = (time() - $timestamp)/60;
                     $wc_order->add_order_note(__("Warning: Payment arrived after $minutes minutes. Received BTC may not match current bitcoin price", 'blockonomics-bitcoin-payments'));
                 }
-                elseif ($status == 2) {
+                elseif ($status >= $network_confirmations) {
```
DarrenWestwood: We should avoid duplicate execution of this code. It may execute on status 0 and then again on status 1 and 2, which will cause multiple unnecessary payment notes.

php/WC_Gateway_Blockonomics.php
freefiona85: Fixed with commit 145b3682e415450e612e3dff6efe184dbf0f9d48

php/WC_Gateway_Blockonomics.php
freefiona85: Fixed with commit 145b3682e415450e612e3dff6efe184dbf0f9d48

freefiona85:
Both fixed with commit 145b3682e415450e612e3dff6efe184dbf0f9d48

Issue #154, opened 2019-10-03
Custom URL routes


jozefpinter:
Hello,

my payment URL looks like

> http://localhost/wc-api/WC_Gateway_Blockonomics/?show_order=1C8965rf2BP44Krtm2imutAXAfgFwt9Lth

and I want something like this

> http://localhost/crypto/?show_order=1C8965rf2BP44Krtm2imutAXAfgFwt9Lth

Replace **/wc-api/WC_Gateway_Blockonomics/** with **/crypto/**. That's all.

My **functions.php** looks like

```
add_action( 'init', 'wpse_rewrites_init' );

function wpse_rewrites_init(){
    add_rewrite_rule(
        'crypto/?show_order=([A-Za-z0-9]+)/?$',
      //'wc-api=WC_Gateway_Blockonomics&show_order=$matches[1]',
        'wc-api=WC_Gateway_Blockonomics&show_order=$1',
        'top' );
}
```

Somebody on the internet suggested that **$matches[1]** doesn't work for external redirects, only internal. But anyways, it doesn't work for me.

I was told to consult this issue with Blockonomics dev Support.

blockonomics:
Thanks for taking out the time to try out our plugin and logging this. We are looking at this on priority

DarrenWestwood:
Hi @jozefpinter 

I am having success using the below code:
```
/* Blockonomics Rewrite Rules */
add_action('init', 'blockonomics_rewrite_rules');
function blockonomics_rewrite_rules() {
    add_rewrite_rule( 'crypto/?$', 'index.php?wc-api=WC_Gateway_Blockonomics', 'top' );
}
```
You will now be able to access the checkout page from 
`http://localhost/crypto/?show_order=1C8965rf2BP44Krtm2imutAXAfgFwt9Lth`

However, more changes will still be required to forward the user to this new URL. I will update you here with this information, when available.

jozefpinter:
Hello @DarrenWestwood ,

exactly as you are saying. With your code I can access payment URL via new URL.

But I need get of rid of old URL because payment gateway is sending me there. So **/wc-api/WC_Gateway_Blockonomics/** is still visible.

DarrenWestwood:
In the callback file [here](https://github.com/blockonomics/woocommerce-plugin/blob/master/php/WC_Gateway_Blockonomics.php#L142), we can add the following line:
[...]

PR #156, opened 2019-10-10
Add 'crypto' custom route issue #154

DarrenWestwood:
Issue #154 - change checkout page URL from 
`http://localhost/wc-api/WC_Gateway_Blockonomics/?show_order=`
to
`http://localhost/crypto/?show_order=`

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -60,6 +60,12 @@ function blockonomics_woocommerce_init()
     add_action('woocommerce_email_customer_details', 'nolo_bnomics_woocommerce_email_customer_details', 10, 1);
     add_filter('woocommerce_payment_gateways', 'woocommerce_add_blockonomics_gateway');
     add_filter('clean_url', 'bnomics_async_scripts', 11, 1 );
+    add_action('init', 'custom_rewrite_rules');
+
+    /* Custom Rewrite Rule */
+    function custom_rewrite_rules() {
+        add_rewrite_rule( 'crypto/?$', 'index.php?wc-api=WC_Gateway_Blockonomics', 'top' );
+    }
 
     /**
      * Add this Gateway to WooCommerce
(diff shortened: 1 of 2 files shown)


php/WC_Gateway_Blockonomics.php
```diff
@@ -139,7 +139,8 @@ public function process_payment($order_id)
         update_option('blockonomics_orders', $blockonomics_orders);
         $order_url = WC()->api_request_url('WC_Gateway_Blockonomics');
         $order_url = add_query_arg('show_order', $address, $order_url);
-
+        $order_url = str_replace('wc-api/WC_Gateway_Blockonomics', 'crypto', $order_url);
```
blockonomics: there is no need to first add WC()->api_request_url and then replace it. We should directly use _show_order_

blockonomics-woocommerce.php
```diff
@@ -60,6 +60,12 @@ function blockonomics_woocommerce_init()
     add_action('woocommerce_email_customer_details', 'nolo_bnomics_woocommerce_email_customer_details', 10, 1);
     add_filter('woocommerce_payment_gateways', 'woocommerce_add_blockonomics_gateway');
     add_filter('clean_url', 'bnomics_async_scripts', 11, 1 );
+    add_action('init', 'custom_rewrite_rules');
+
+    /* Custom Rewrite Rule */
+    function custom_rewrite_rules() {
```
blockonomics: Let us  _show_order_ . Also don't hardcode `wc-api=WC_Gateway_Blockonomics` let us use `WC()->api_request_url` else it won't work for WC < 3.0

blockonomics:
Does this require customer to manually flush rewrite rules or it will auto work on installation / upgrade ?

DarrenWestwood:
> Does this require customer to manually flush rewrite rules or it will auto work on installation / upgrade ?

Yes this does require a manual flush or flush_rewrite_rules() can be added to the installation / upgrade.

blockonomics:
Please fix conflicts and other pull request comments

blockonomics-woocommerce.php
DarrenWestwood: We cannot use `WC()->api_request_url` because add_rewrite_rule requires the url to start with `index.php`. All wordpress urls can be accessed in this manner.
In later versions of woocommerce, they added their own rewrite rules to change the url, however, we still need to access the original page with the form `index.php?wc-api=WC_Gateway_Blockonomics`

DarrenWestwood:
We can also possibly use the URL structure:
`http://localhost/show_order/<address>`

Using `str_replace` will cause issues with users who do not have rewrites activated. They will be forwarded to the new URL which will not exist.
eg. Wordpress > Settings > Permalinks > Select Plain and save

PR #159, opened 2019-10-16
How to Pay Link added + Restyling

freefiona85:
Restyle some font size, and added link to how to pay.

### DIFF
--- a/css/order.css
+++ b/css/order.css
@@ -115,12 +115,11 @@
 
 .bnomics-how-to-pay {
 text-align: center;
-font-size: 12px;
+font-size: 0.8em;
 padding: 0;
 }
 
 .bnomics-how-to-pay a {
-color: #6d6d6d;
 text-decoration: underline;
 font-style: normal;
 }
@@ -129,7 +128,7 @@ font-style: normal;
   text-align: center;
   margin-top: 1.6em;
   margin-bottom: 1em;
-  font-size: 12px;
+  font-size: 0.8em;
   padding: 0 20px;
 }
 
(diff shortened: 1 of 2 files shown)


css/order.css
```diff
@@ -108,16 +108,28 @@
...
 }
 
+.bnomics-how-to-pay {
+text-align: center;
+font-size: 12px;
+padding: 0;
+}
+
+.bnomics-how-to-pay a {
+color: #6d6d6d;
```
blockonomics: We shouldn't use the color style. Our main focus is to respect the website theme
Actually we should also not use absolute font-size and only use <h1> <h2> etc..

But we are already using font-size is various places (I am not sure why ?). Maybe we could switch to relative font-size (so if we only specify percentage change relative to theme's default font size). @DarrenWestwood can give more detail on this

freefiona85: ok, so, color style for the link should not be done, too? because a link would be different. also, we can use em measurements for the fontsize if it's okay.

blockonomics: yes. lets use em everywhere. Please attach a screenshot on how it would look as a normal link

freefiona85: image
I already removed the color too.

blockonomics: Yes, the hyperlink color is not looking good. Is there a way to force to the default color of the theme instead of picking up the `<a>` tag color

freefiona85: i used color:inherit. it seems to be working.
image

blockonomics: yes this looks good. @DarrenWestwood  please review the pull

freefiona85:
> Changes integrate well on multiple themes tested:
> WordpressTheme2
> 
> We should hide the how to pay link from the expired page:
> Wordpress

Updated with commit #08042ab - it won't appear on expiry page.

PR #160, opened 2019-10-17
Added refresh on expiry

freefiona85:
Refresh keeps using same order id, same address, only resetting timer (clock). Question, should I reset the timestamp as well?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+
+assets-wp-repo/Thumbs.db
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,0 @@
-
-assets-wp-repo/Thumbs.db
--- a/js/app.js
+++ b/js/app.js
@@ -78,7 +78,6 @@ function getParameterByNameBlocko(name, url) {
 app.controller('CheckoutController', function($scope, $interval, Order, $httpParamSerializer, $timeout, AltcoinNew, AltcoinAccept, AltcoinLimits, WpAjax) {
     //get order id from url
     $scope.address = getParameterByNameBlocko("show_order");
-	$scope.refresh = getParameterByNameBlocko("refresh");
     var totalProgress = 100;
     $scope.copyshow = false;
     //blockonomics_time_period is defined on JS file as global var
@@ -232,10 +231,6 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
                 //Mark order as expired if we ran out of time
                 if ($scope.clock < 0) {
 						$scope.order.status = -3;
-						if($scope.refresh == 1) {
-							$scope.order.status = -1;
-							$scope.clock = totalTime;
-						} else { 
 						return;
 						}
                 }
--- a/js/app.js
+++ b/js/app.js
@@ -78,6 +78,7 @@ function getParameterByNameBlocko(name, url) {
 app.controller('CheckoutController', function($scope, $interval, Order, $httpParamSerializer, $timeout, AltcoinNew, AltcoinAccept, AltcoinLimits, WpAjax) {
     //get order id from url
     $scope.address = getParameterByNameBlocko("show_order");
+	$scope.refresh = getParameterByNameBlocko("refresh");
     var totalProgress = 100;
     $scope.copyshow = false;
     //blockonomics_time_period is defined on JS file as global var
@@ -230,8 +231,13 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
                 $scope.clock = $scope.order.timestamp + totalTime - Math.floor(Date.now() / 1000);
                 //Mark order as expired if we ran out of time
                 if ($scope.clock < 0) {
-                    $scope.order.status = -3;
-                    return;
+						$scope.order.status = -3;
+						if($scope.refresh == 1) {
+							$scope.order.status = -1;
+							$scope.clock = totalTime;
+						} else { 
+						return;
+						}
                 }
                 $scope.tick_interval = $interval($scope.tick, 1000);
                 //Connect and Listen on websocket for payment notification
--- a/js/app.js
+++ b/js/app.js
@@ -9,11 +9,11 @@ service.factory('Order', function($resource) {
         param = {
             "wc-api": param
         };
-    else
-        param = {};
-    var item = $resource(window.location.pathname, param);
-    return item;
-});
+        else
+            param = {};
+        var item = $resource(window.location.pathname, param);
+        return item;
+    });
 
 var flyp_base = 'https://flyp.me/api/v1';
 
@@ -68,7 +68,7 @@ function getParameterByNameBlocko(name, url) {
     }
     name = name.replace(/[\[\]]/g, "\\$&");
     var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)"),
-        results = regex.exec(url);
+    results = regex.exec(url);
     if (!results) return null;
     if (!results[2]) return '';
     return decodeURIComponent(results[2].replace(/\+/g, " "));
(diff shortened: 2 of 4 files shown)


js/app.js
```diff
@@ -230,8 +231,13 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
                 $scope.clock = $scope.order.timestamp + totalTime - Math.floor(Date.now() / 1000);
                 //Mark order as expired if we ran out of time
                 if ($scope.clock < 0) {
-                    $scope.order.status = -3;
-                    return;
+						$scope.order.status = -3;
```
DarrenWestwood: line indents can be adjusted correctly

templates/order.php
```diff
@@ -54,7 +54,7 @@
                 <!-- Order Status -->
                 <div class="bnomics-order-status-wrapper">
                   <span class="bnomics-order-status-title" ng-show="order.status == -1" ng-cloak ><?=__('To confirm your order, please send the exact amount of <strong>BTC</strong> to the given address', 'blockonomics-bitcoin-payments')?></span>
-                  <span class="warning bnomics-status-warning" ng-show="order.status == -3" ng-cloak><?=__('Payment Expired (Use the browser back button and try again)', 'blockonomics-bitcoin-payments')?></span>
+                  <span class="warning bnomics-status-warning" ng-show="order.status == -3" ng-cloak><?=__('<b>PAYMENT EXPIRED</b> <br /><br /><ul><li><div style="width:100%;display:table;"><span style="width:20%;display:table-cell;"><b>If you already paid :</b></span><span style="display:table-cell;"> Order will automatically confirmed when we detect your payment. You can safely close this window.</span></div></li><li> <b>If you have not already made the payment :</b> <a href="?show_order={{order.address}}&refresh=1">Click here</a> to try again.</li></ul>', 'blockonomics-bitcoin-payments')?></span>
```
DarrenWestwood: Can we simplify the UI. Possibly something like this:
Wordpress Payment Expired Update

templates/order.php
```diff
@@ -80,6 +80,9 @@
                 <span class="ng-cloak bnomics-time-left" ng-hide="order.status != -1">{{clock*1000 | date:'mm:ss' : 'UTC'}} min left to pay your order</span>
               </div>
         <!-- Blockonomics Credit -->
+			<div class="bnomics-how-to-pay">
```
DarrenWestwood: Let's remove the changes from the how-to-pay branch.

js/app.js
```diff
@@ -78,6 +78,7 @@ function getParameterByNameBlocko(name, url) {
 app.controller('CheckoutController', function($scope, $interval, Order, $httpParamSerializer, $timeout, AltcoinNew, AltcoinAccept, AltcoinLimits, WpAjax) {
     //get order id from url
     $scope.address = getParameterByNameBlocko("show_order");
+	$scope.refresh = getParameterByNameBlocko("refresh");
```
blockonomics: I don't think a refersh parameter is needed. This causes a problem because if customer presses F5 on payment screen. It will again restart the payment timer, which is not a desired behaviour.

What we need to do is to do a call to backend to update order of timestamp/BTC value of order (according to latest price) and then just reload the *show_order* page. It should automatically work

templates/order.php
freefiona85: done.

templates/order.php
freefiona85: changed accordingly.

js/app.js
freefiona85: Modified with the newest commit to the branch.

DarrenWestwood: We should remove the reset parameter from the url as this will cause a new address to be generated on consecutive refreshes.  I recommend using the timestamp to check if an order is expired and generate a new timestamp and BTC value if required.

templates/order.php
```diff
@@ -54,7 +54,7 @@
                 <!-- Order Status -->
                 <div class="bnomics-order-status-wrapper">
                   <span class="bnomics-order-status-title" ng-show="order.status == -1" ng-cloak ><?=__('To confirm your order, please send the exact amount of <strong>BTC</strong> to the given address', 'blockonomics-bitcoin-payments')?></span>
-                  <span class="warning bnomics-status-warning" ng-show="order.status == -3" ng-cloak><?=__('Payment Expired (Use the browser back button and try again)', 'blockonomics-bitcoin-payments')?></span>
+                  <span class="warning bnomics-status-warning" ng-show="order.status == -3" ng-cloak><?=__('<b>PAYMENT EXPIRED</b> <br /><br /><a href="/index.php/checkout?reset=1">Click here</a> to try again.<br /><br /><p>If you already paid, your order will be processed automatically. <br />You can safely close this window.</p>', 'blockonomics-bitcoin-payments')?></span>
```
DarrenWestwood: `/index.php/checkout?reset=1` is breaking websites.
Screenshot from 2019-10-28 02-14-46

DarrenWestwood: There are display differences in the text CSS. 
Different text formatting

freefiona85: Fixed as we use div now instead of <p>.

freefiona85: https://github.com/blockonomics/woocommerce-plugin/pull/160#discussion_r339374488
fixing it using wss call to get timestamp. will be there on next commit.

js/app.js
freefiona85: reset parameter, as per Blockonomics API, won't generate new address (it's what we sent to newAddress function) as in the 'pay with bitcoin' page we didn't have access to the current address. And as process_payment is only used when we click 'pay with bitcoins' function (overridden from woocommerce i assume), that's where we should do the check. 

Should i make another option to save up current address? Timestamp call needs address too.

freefiona85: OK, pushed new commit. This time using timestamp for checking, and order_id as index from blockonomics_orders.

php/WC_Gateway_Blockonomics.php
```diff
@@ -108,7 +108,15 @@ public function process_payment($order_id)
         $cancel_url = add_query_arg('order_key', $order->get_order_key(), $cancel_url);
 
         $blockonomics = new Blockonomics;
-        $responseObj = $blockonomics->new_address(get_option("blockonomics_callback_secret"));
+		$time_period = get_option("blockonomics_timeperiod", 10) *60;
+		$currentOrder = $blockonomics_orders[$order_id];
+		if(isset($currentOrder[$address])) {
+			if(time() > $timestamp + $time_period) {
+				$responseObj = $blockonomics->new_address(get_option("blockonomics_callback_secret"),1);
```
DarrenWestwood: Using reset=1 will return the last generated address for the wallet. This needs to instead reuse the previous address linked to the order.

freefiona85:
@DarrenWestwood fixed using order_id checking, so if the order_id has the address, it will use that address instead.

php/WC_Gateway_Blockonomics.php
freefiona85: removed all the parameters, and reverted from using the reset parameter.

.gitignore
```diff
@@ -0,0 +1,2 @@
+
+assets-wp-repo/Thumbs.db
```
DarrenWestwood: This .gitignore file can be removed from the pr

js/app.js
```diff
@@ -230,8 +230,9 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
                 $scope.clock = $scope.order.timestamp + totalTime - Math.floor(Date.now() / 1000);
                 //Mark order as expired if we ran out of time
                 if ($scope.clock < 0) {
-                    $scope.order.status = -3;
-                    return;
+						$scope.order.status = -3;
+						return;
+						}
```
DarrenWestwood: This line has been included as an error. Using a text editor like [Sublime Text](https://www.sublimetext.com), can avoid such issues and correct the issues with spacing and indents in your code.

.gitignore
freefiona85: ok, both done with latest commit.

js/app.js
freefiona85: ok, done.

js/app.js
```diff
@@ -230,8 +230,9 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
                 $scope.clock = $scope.order.timestamp + totalTime - Math.floor(Date.now() / 1000);
                 //Mark order as expired if we ran out of time
                 if ($scope.clock < 0) {
-                    $scope.order.status = -3;
-                    return;
+					$scope.order.status = -3;
+					return;
+					}
```
DarrenWestwood: 67731774-ebd49680-fa01-11e9-8f82-6ed087871487
This error is being caused by the extra `}` you have included. Please check your opening and closing tags. This should be removed as it is breaking the checkout on my end.

If you visit https://github.com/blockonomics/woocommerce-plugin/pull/160/files, you will notice all of your new code is shifted further right than the existing code. Please correct this.

DarrenWestwood:
@freefiona85 , $time_period can also be removed as it does not get used. The indents could still further be improved. Your text editor is using tab characters whereas the file currently uses spaces.
Sublime

DarrenWestwood:
@blockonomics , The changes are functioning as desired.
- Pay with bitcoin: For same order always show same address. Clicking back and then trying this again will then effectively refresh the timer and price
- Refresh F5 on payment page: Should not reset timer. Should continue on timestamp before refersh

Issue #164, opened 2019-11-07
Already Paid ? 


blockonomics:
Click here to go back it creating issues. We can remove that and replace it with 

**Already Paid ? Click here** > *on clicking expands*
We haven't detected your payment yet. Please wait a while for your transaction to confirm. If it is already confirmed, there maybe problem with paid amount, **Click here** to get refund (*goes to our existing refund screen*)

PR #165, opened 2019-11-21
Change behaviour from navigation from payment page to showing more in…

jusasiiv:
…formation and option for refund. Issue #164


blockonomics:
Functionality is as expected

Issue #172, opened 2020-03-27
Personal payments causing callback error


shivaenigma:
If callback address is not a part of order, it should fail gracefully

![Screenshot from 2020-03-27 17-42-06](https://user-images.githubusercontent.com/737092/77757405-3280f800-7057-11ea-8f92-8518b4a1e26a.png)

shivaenigma:
Still is still not fixed. Found this on merchant store having plugin version 1.8.3 
![Screenshot_2020-06-01 WordPress › fout](https://user-images.githubusercontent.com/737092/83393544-98517d80-a414-11ea-884b-1e1520d8e96a.png)

PR #173, opened 2020-04-10
Handle order not found, #172

DarrenWestwood:
Handle error in case order not found by address, #172

### DIFF
--- a/php/WC_Gateway_Blockonomics.php
+++ b/php/WC_Gateway_Blockonomics.php
@@ -194,9 +194,9 @@ class WC_Gateway_Blockonomics extends WC_Payment_Gateway
         $secret = isset($_REQUEST['secret']) ? $_REQUEST['secret'] : "";
         $network_confirmations=get_option("blockonomics_network_confirmation",2);
         if ($callback_secret  && $callback_secret == $secret) {
-            $addr = $_REQUEST['addr'];
-            $order = $orders[$addr];
-            if ($order) {
+            $addr = isset($_REQUEST['addr']) ? $_REQUEST['addr'] : "";
+            if (array_key_exists($addr, $orders)){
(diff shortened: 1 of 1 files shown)


php/WC_Gateway_Blockonomics.php
```diff
@@ -196,8 +196,8 @@ public function check_blockonomics_callback()
         if ($callback_secret  && $callback_secret == $secret) {
             $addr = $_REQUEST['addr'];
             $order = $orders[$addr];
```
shivaenigma, line 198: Giving undefined index on when looking up addr inside orders
`Notice: Undefined index: 18MKYd8YnVtsAakrTmpBm3qLfsXLdKjsqf in /home/enigma/blockonomics-bitcoin-payments/php/WC_Gateway_Blockonomics.php on line 198`

php/WC_Gateway_Blockonomics.php
```diff
@@ -241,7 +241,11 @@ public function check_blockonomics_callback()
                     update_post_meta($wc_order->get_id(), 'expected_btc_amount', $order['satoshi']/1.0e8);
                 }
                 update_option('blockonomics_orders', $orders);
+            }else{
+                exit("Error: order not found");
             }
+        }else{
+            exit();
```
shivaenigma, line 248: else and exit() here is uncessary as the function is anyway existing

php/WC_Gateway_Blockonomics.php
DarrenWestwood: Updated to check key exists - 2d63e48

php/WC_Gateway_Blockonomics.php
DarrenWestwood: Removed in 3a92048

PR #180, opened 2020-05-25
BCH UI and settings integration

badr-r:


DarrenWestwood:
Code changes are looking good. You will need to store the currency uri (eg. bitcoin, bitcoincash) with the currency code (btc, bch). This is required for the qrcode to also switch to `bitcoincash` appropriately depending on the selected currency:
https://github.com/badr-r/woocommerce-plugin/blob/master/templates/blockonomics_checkout.php#L65

Some example code:
https://github.com/blockonomics/whmcs-bitcoin-plugin/blob/master/modules/gateways/Blockonomics/Blockonomics.php#L97-L111

js/app.js
```diff
@@ -100,24 +66,23 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
-            }
-            return url;
+    
+    // Select Blockonomics currency
+    $scope.select_blockonomics_currency = function(blockonomics_currency) {
+        $scope.currency = blockonomics_currency;
+        $scope.currency_selecter  = false;
+    }
+
+    Order.get({
```
DarrenWestwood, line 76: This can be done before loading the page and then passing the active currencies to the scope using something like this: 
```
<div id="active_currencies" data-active_currencies='<?php echo $active_currencies; ?>'></div>
```
and
```
    var active_currencies = JSON.parse(active_currencies_div.dataset.active_currencies);
    $scope.active_currencies = active_currencies;
```
This will save on an additional call from the client side to server side.

DarrenWestwood:
The recent changes are better but it is throwing an error in the console:
`Error: $scope.active_currencies.bch is undefined`. This is because we are using `get_option('blockonomics_active_currency')`, which does not exist. 
Lets rather use `json_encode($blockonomics->getActiveCurrencies())` to fetch the active currencies.

css/order.css
```diff
@@ -340,4 +340,73 @@ color:inherit;
...
+  cursor:pointer;
+}
+
+.bnomics-select-options h3 {
+  display:inline;
+  margin: 0;
+}
+
+.bnomics-select-container table{
+  border-spacing: 0 1em;
```
DarrenWestwood, line 383: Some themes still showing the border around the table. Lets use `border: 0;` to remove the border.

PR #183, opened 2020-06-09
Bchintegration

jusasiiv:


php/Blockonomics.php
```diff
@@ -29,7 +32,7 @@ public function get_api_key()
     }
 
 
-    public function new_address($secret, $reset=false)
+    public function new_address($secret, $crypto,$reset=false)
```
jusasiiv, line 35: See a call to this function from line 271 in `test_setup()` Function params changed but the call did not so its not actually doing reset but passing $crypto as Ture and reset defaults to false

jusasiiv, line 35: See a call to this function from line 271 in `test_setup()` Function params changed but the call did not so its not actually doing reset but passing `$crypto` as True and reset defaults to false

php/Blockonomics.php
```diff
@@ -84,6 +95,37 @@ public function get_temp_api_key($callback_url)
...
+                    'name' => 'Bitcoin Cash',
+                    'uri' => 'bitcoincash',
+                    'enabled' => ''
+              )
+          );
+    }
+
+	//Get list of active crypto currencies
+	public function getActiveCurrencies() {
+		// $active_currencies = array();
```
jusasiiv, line 116: Lets not leave code in comments

php/Blockonomics.php
```diff
@@ -84,6 +95,37 @@ public function get_temp_api_key($callback_url)
...
+        $active_currencies = $this->getSupportedCurrencies();
+		foreach ($active_currencies as $code => $currency) {
```
jusasiiv, line 118: This loop is hard to follow. We could clarify it by doing `if code == btc` > `continue` in the beginning of the loop

`getSupportedCurrencies` array is mixing strings and boolean for enabled. By setting enabled to false for other than BTC we can also drop more from this loop

Also dont mix tabs and spaces for indentation

php/WC_Gateway_Blockonomics.php
```diff
@@ -172,64 +134,106 @@ public function check_blockonomics_callback()
...
-            exit(json_encode($orders[$address]));
+            exit(json_encode($orders[$order_id]));
         }
 
         $callback_secret = get_option("blockonomics_callback_secret");
         $secret = isset($_REQUEST['secret']) ? $_REQUEST['secret'] : "";
         $network_confirmations=get_option("blockonomics_network_confirmation",2);
         if ($callback_secret  && $callback_secret == $secret) {
             $addr = isset($_REQUEST['addr']) ? $_REQUEST['addr'] : "";
-            if (array_key_exists($addr, $orders)){
```
jusasiiv, line 206: Why was this code changed? No need to revert but comment why this change is necessary. To my eye it looks like it is doing the same thing but more complexly

php/WC_Gateway_Blockonomics.php
```diff
@@ -245,27 +249,13 @@ public function check_blockonomics_callback()
...
+                    update_post_meta($wc_order->get_id(), 'expected_'. $order['crypto'] .'_amount', $order['satoshi']/1.0e8);
                 }
                 update_option('blockonomics_orders', $orders);
             }else{
                 exit("Error: order not found");
             }
         }
     }
-
-    private function displayError($woocommerce) {
```
jusasiiv, line 258: Just a question but why was this moved to template?

php/WC_Gateway_Blockonomics.php
```diff
@@ -108,42 +104,8 @@ public function process_payment($order_id)
         $order_key = method_exists( $order, 'get_order_key' ) ? $order->get_order_key() : $order->order_key;
         $cancel_url = add_query_arg('order_key', $order_key, $cancel_url);
 
-        $blockonomics = new Blockonomics;
```
jusasiiv, line 111: Also just a question to understand better, what was the reason of moving this code?

templates/blockonomics_checkout.php
```diff
@@ -95,35 +114,27 @@
...
-                </div>
-            </div>
-          </div>
-          <?php endif ?>
-
         </div>
       </div>
+    </div>
+      <!-- Address Error -->
+      <div class="bnomics-display-error" style="text-align: center;" ng-show="btcaddresserror">
```
jusasiiv, line 121: Two different designs on error style is not needed. One div is enough and only change which error text is displayed

php/WC_Gateway_Blockonomics.php
jusasiiv, line 206: This change is doing same thing as previous. Its cleaner to use `array_key_exists()` than to implement the same thing in a `for each` loop

js/app.js
```diff
@@ -92,32 +54,79 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
-            url = window.location.pathname;
-            var serializedParams = $httpParamSerializer(params);
-            if (serializedParams.length > 0) {
-                url += ((url.indexOf('?') === -1) ? '?' : '&') + serializedParams;
-            }
-            return url;
+    
+    // Select Blockonomics currency
+    $scope.select_blockonomics_currency = function(blockonomics_currency) {
+        $scope.currency = blockonomics_currency;
```
jusasiiv, line 68: The value is assigned to both `$scope.currency` and `blockonomics_currency;` and both are used in the code. I guess `$scope.currency`is used in frontend somewhere as its assigned to scope. Cleaner to use `$scope.currency` in all parts of code

js/app.js
```diff
@@ -92,32 +54,79 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
+                "get_order": $scope.order_id,
+                'crypto': blockonomics_currency
+            }, function(data) {
+                $scope.order = data;
+                //Check the status of the order
+                if ($scope.order.status == -1) {
+                  $scope.clock = $scope.order.timestamp + totalTime - Math.floor(Date.now() / 1000);
+                    //Mark order as expired if we ran out of time
+                    if ($scope.clock < 0) {
+                    $scope.order.status = -3;
```
jusasiiv, line 84: Minor thing but as JS tends to get messy its good practice to intend correctly, same goes for other lines

js/app.js
```diff
@@ -260,263 +152,44 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
+        textarea.style.height = 0;
+        document.body.appendChild(textarea);
+        textarea.value = document.getElementById("bnomics-amount-copy").innerText;
+
+        var selector = document.querySelector('#temp_element');
+        selector.select();
+        document.execCommand('copy');
+        document.body.removeChild(textarea);
+        select_text("bnomics-amount-copy");
+        // var copyText = document.getElementById("bnomics-amount-copy");
```
jusasiiv, line 168: Code in comments

PR #187, opened 2020-06-29
bch integration + refactor

shivaenigma:
We need to organize and simplify code in a final pass before merging in bch integration

- currency selection will be in a new template *blockonomics_crypto_options.php*
- *show_order=id&crypto=btc/bch* will display checkout page. *select_crypto_order=id* will show *blockonomics_crypto_options.php* . All address generation/reuse happens in checkout page. *blockonomics_crypto_options.php*  is only for currency selection
- If merchant only enabled BTC ..on checkout it directly does to checkout page  
- *check_blockonomics_callback* function has become very monoltihic. Refactor it into smaller functions. Also there are no code comments. Add code comments at appropriate places
- Refresh of checkout page should work. It shouldn't mess up callback logic and timer shouldn't reset on refresh (same as existing functionality) 
- Make  sure all functionality/error messages are correct 

Bugs/Cleanup
- https://github.com/blockonomics/woocommerce-plugin/blob/bchintegration/php/WC_Gateway_Blockonomics.php#L250 does this make sense ? We are now indexing using order_id right ?
- https://github.com/blockonomics/woocommerce-plugin/blob/bchintegration/php/WC_Gateway_Blockonomics.php#L105. It can't find the usage for *cancel_url* and few above lines. Remove if not used
- Remove other old code

Important
- Thoroughly test old functionality / BTC checkout. It is very vital that none of the old functionality is affected as we have a large userbase


DarrenWestwood:
- Fixes issue overwriting addresses in blockonomics_orders
- Updated to the new design
- Code refactor and improvements
- Existing functionality tested

shivaenigma:
- Remove https://prnt.sc/tpqel1 from crypto select template
- Clicking on back from payment screen, the crypto select screen is showing just a spinner. Only manually refreshing the page is working https://drive.google.com/file/d/1HZaOlFM0lbAC6jiu-tUzaV_SF_evN0j8/view?usp=sharing

DarrenWestwood:
> Remove https://prnt.sc/tpqel1 from crypto select template

Removed in 7cd7df9

> Clicking on back from payment screen, the crypto select screen is showing just a spinner.

Fixed in 778d07a. @shivaenigma, please could you test on your end.

shivaenigma:
I can still see back button spinner issue. I am using theme twenty seventeen

Can you explain 778d07a68ad230. It can't quite understand the motivation behind it. The commit message doesn't describe the changes in the commit

DarrenWestwood:
778d07a was added as I suspected using window instead of injecting $window may have been causing this behaviour: https://docs.angularjs.org/api/ng/service/$window

shivaenigma:
Getting this in debug mode
Screenshot from 2020-08-11 18-15-13

templates/blockonomics_checkout.php
```diff
@@ -1,144 +1,103 @@
...
-  <link rel="stylesheet" type="text/css" href="<?php echo plugins_url('css/icons/icons.css', dirname(__FILE__));?>">
-<?php
-}else{
-  get_header();
-}
-?>
-<div ng-app="shopping-cart-demo">
+<div id="time_period" data-time_period="<?php echo get_option('blockonomics_timeperiod', 10); ?>"></div>
+<div id="active_cryptos" data-active_cryptos='<?php echo json_encode(get_option('blockonomics_active_cryptos')); ?>'></div>
+
```
shivaenigma, line 3: This is unused. Please remove

blockonomics-woocommerce.php
```diff
@@ -97,6 +97,7 @@ function add_page()
     {
         include_once plugin_dir_path(__FILE__) . 'php' . DIRECTORY_SEPARATOR . 'Blockonomics.php';
         $blockonomics = new Blockonomics;
+        update_option('blockonomics_active_cryptos',$blockonomics->getActiveCurrencies());
```
shivaenigma, line 100: Adding a new option *blockonomics_active_cryptos* is a bad design as it is duplicating data. This information can be easily fetched via *getActiveCurrencies()* which is a very cheap function call. Remove this option.

js/app.js
```diff
@@ -1,123 +1,92 @@
...
+function CryptoOptionsController($scope, $httpParamSerializer) {
+    var active_cryptos_div = document.getElementById("active_cryptos");
+    var active_cryptos = JSON.parse(active_cryptos_div.dataset.active_cryptos);
+    $scope.no_display_error = true;
+    $scope.active_cryptos = active_cryptos;
+    $scope.crypto_selecter  = true;
+    //fetch url params
+    $scope.order_id = getParameterByNameBlocko("select_crypto");
+
+    $scope.checkout_order_url = function() {
```
shivaenigma, line 25: There is visible code duplication in *checkout_order_url* and *finish_order_url* methods. Re factor them to a single method. *get_wc_endpoint* .. it takes dict parameters are input and returns a wc URL. No need to use $scope variable inside these methods.

js/app.js
```diff
@@ -1,123 +1,92 @@
...
 
-function getParameterByNameBlocko(name, url) {
-    if (!url) {
-        url = window.location.href;
+    //Select Blockonomics crypto
+    $scope.select_blockonomics_crypto = function(blockonomics_crypto) {
+        $scope.crypto_selecter  = false;
+        $scope.spinner = true;
+        $scope.crypto = $scope.active_cryptos[blockonomics_crypto];
+        $scope.crypto.code = blockonomics_crypto;
```
shivaenigma, line 48: Can't find the use of $scope.crypto inside CryptoOptionsController.  After we refactor *checkout_order_url* we just pass it order_id and *blockonomics_crypto* to that function

js/app.js
```diff
@@ -131,392 +100,140 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
-            return true;
-        }
-        return false;
-    }
-    function isRefundAddress(){
-        if( 'refund_address' in data ){
-            return true;
-        }
-        return false;
+    //Copy bitcoin address to clipboard
```
shivaenigma, line 205: Incorrect/Misleading comment

js/app.js
```diff
@@ -131,392 +100,140 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
-
-    //Go to add refund page
-    $scope.get_refund = function() {
-        stop_interval();
-        $scope.hide_refund_reason = true;
-        update_altcoin_status('add_refund');
-    }
+Order.$inject = ['$resource'];
+function Order($resource) {
+    //There are two styles of callback url in
```
shivaenigma, line 213: This should also use the new suggested function *get_wc_endpoint*. Move these code comments inside *get_wc_endpoint* so that we better understand the need of that function

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+        add_action('wp_enqueue_scripts', 'bnomics_enqueue_stylesheets' );
+        // Apply nojs header changes
+        if (!$this->is_nojs_template($template_name)) {
+            add_action('wp_enqueue_scripts', 'bnomics_enqueue_scripts' );
+        }
+        // Apply lite-mode header
+        if($this->is_lite_mode_active()){
+        ?>
+          <link rel="stylesheet" type="text/css" href="<?php echo plugins_url('css/order.css', dirname(__FILE__));?>">
+        <?php
```
shivaenigma, line 337: Why is this necessary when bnomics_enqueue_stylesheets is already addng order.css ?

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+    }
+
+    // Adds the header to the blockonomics page
+    public function load_blockonomics_header($template_name){
+        add_action('wp_enqueue_scripts', 'bnomics_enqueue_stylesheets' );
+        // Apply nojs header changes
+        if (!$this->is_nojs_template($template_name)) {
+            add_action('wp_enqueue_scripts', 'bnomics_enqueue_scripts' );
+        }
+        // Apply lite-mode header
```
shivaenigma, line 333: Comments shouldn't describe what is evident from code. It should describe motivation/more information.For example here coment could be:

**Lite mode will render without wordpress theme headers**

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+        }else{
+          get_header();
+        }
+    }
+
+    // Adds the footer to the blockonomics page
+    public function load_blockonomics_footer($template_name){
+        // Apply lite-mode footer changes
+        if($this->is_lite_mode_active()){
+            // Apply nojs footer changes
```
shivaenigma, line 347: Code Comments should be descriptive and not be self evident

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+            load_template( plugin_dir_path(__FILE__)."../templates/" .$template );
+        }
+
+        $this->load_blockonomics_footer($template_name);
+
+        exit();
+    }
+
+    // Load the the payment confirmed template in the page
+    public function load_payment_confirmation_template(){
```
shivaenigma, line 384: Rename function name that indicates it is for noJS flow

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+        }
+        $wc_order = new WC_Order($order_id);
+        $order = $orders[$order_id][$address];
+        // Check if order has not expired yet
+        if ( isset($order['timestamp']) && $order['timestamp'] >= time() - get_option("blockonomics_timeperiod") * 60 ) {
+            $timestamp = $order['timestamp'];
+            $satoshi = $order['satoshi'];
+            $status = $order['status'];
+        }else{
+            // Refresh the timestamp and expected satoshi
```
shivaenigma, line 455: This also handles new orders, so the comment in misleading

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+        if ($callback_secret  && $callback_secret == $secret) {
+            $orders = get_option('blockonomics_orders');
+            // fetch the order id by bitcoin address
+            foreach($orders as $id => $order){
+                if(isset($order[$addr])){
+                    $order_id = $id;
+                    break;
+                }
+            }
+            if ($order_id){
```
shivaenigma, line 507: Lots of nested ifs cause confusion in code. A better thing to do here is 
```
if (!isset($order_id))
exit("Error: order not found");
```
and carry on forward without adding an extra if nesting

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+                $order = $orders[$order_id][$addr];
+                $wc_order = new WC_Order($order_id);
+                $existing_status = $order['status'];
+                $timestamp = $order['timestamp'];
+                $time_period = get_option("blockonomics_timeperiod", 10) *60;
+                $network_confirmations=get_option("blockonomics_network_confirmation",2);
+                if ($status == 0 && time() > $timestamp + $time_period) {
+                    $minutes = (time() - $timestamp)/60;
+                    $wc_order->add_order_note(__("Warning: Payment arrived after $minutes minutes. Received". $order['crypto'] ."may not match current ". $order['crypto'] ." price", 'blockonomics-bitcoin-payments'));
```
shivaenigma, line 516: Remove this. It is confusing and uncessary

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+                }
+                elseif ($status >= $network_confirmations && !metadata_exists('post',$wc_order->get_id(),'paid_'. $order['crypto'] .'_amount') )  {
+                    update_post_meta($wc_order->get_id(), 'paid_'. $order['crypto'] .'_amount', $value/1.0e8);
```
shivaenigma, line 519: This is a bit confusing on why were are checking metadata etc. Create a method

*recordPayment()* : Params value, order_id and cryto. If payment is already recorded returns False else records payment and return True.

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+                        update_option('blockonomics_temp_withdraw_amount', $new_temp_amount);
+                    }
+                }
+                $order['txid'] = $txid;
+                $order['status'] = $status;
+                $orders[$order_id][$addr] = $order;
+                if ($existing_status == -1) {
+                    update_post_meta($wc_order->get_id(), 'blockonomics_'. $order['crypto'] .'_txid', $order['txid']);
+                    update_post_meta($wc_order->get_id(), 'expected_'. $order['crypto'] .'_amount', $order['satoshi']/1.0e8);
+                }
```
shivaenigma, line 551: This extra condition is necessary now that we are checking for paid_crypto metadata

php/Blockonomics.php
```diff
@@ -238,4 +292,267 @@ public function testSetup()
...
+    public function is_payment_pending($order_id){
+        $orders = get_option('blockonomics_orders');
+        $network_confirmations = get_option("blockonomics_network_confirmation",2);
+        foreach ($orders[$order_id] as $addr => $order){
+            if ($order['status'] >= 0 && $order['status'] < $network_confirmations){
+                return true;
+            };
+        };
+        return false;
+    }
```
shivaenigma, line 398: Remove this method and call invocations.
Modify process_order() so that it just returns order_info when existing status!=-1 (Doesn't change order information)

blockonomics-woocommerce.php
DarrenWestwood: Removed in b786b57 and b6945c4

js/app.js
DarrenWestwood: Refactored in 1f5a634

js/app.js
DarrenWestwood: Fixed in cd0b822

js/app.js
DarrenWestwood: Fixed in a92d16c

js/app.js
DarrenWestwood: Refactored in 130cef6

php/Blockonomics.php
DarrenWestwood: This is necessary for lite mode as enqueued stylesheets are loaded in the wordpress header, which does not get loaded.

php/Blockonomics.php
DarrenWestwood: Fixed in 105539b

php/Blockonomics.php
DarrenWestwood: Fixed in 105539b

php/Blockonomics.php
DarrenWestwood: Renamed in 9261a5c

php/Blockonomics.php
DarrenWestwood: Fixed in 9e2864d

php/Blockonomics.php
DarrenWestwood: Fixed in 7f904d4

php/Blockonomics.php
DarrenWestwood: This is still required as it is the only way for merchants to see if a payment was sent after the timer has expired. Without this check, payments can still be sent at a later stage with the incorrect conversion and merchants will not know. Refactored in 62c1ec2

php/Blockonomics.php
DarrenWestwood: Refactored in 62c1ec2

php/Blockonomics.php
DarrenWestwood: Refactored in 62c1ec2

php/Blockonomics.php
DarrenWestwood: Fixed in 7f904d4

blockonomics-woocommerce.php
```diff
@@ -502,11 +454,12 @@ function blockonomics_uninstall_hook() {
     delete_option('blockonomics_margin');
     delete_option('blockonomics_timeperiod');
     delete_option('blockonomics_api_updated');
-    delete_option('blockonomics_altcoins');
+    delete_option('blockonomics_bch');
     delete_option('blockonomics_underpayment_slack');
     delete_option('blockonomics_lite');
     delete_option('blockonomics_nojs');
 	delete_option('blockonomics_network_confirmation');
+	delete_option('blockonomics_active_cryptos');
```
shivaenigma, line 462: Remove unused option

js/app.js
```diff
@@ -131,392 +58,160 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
+                if(data.address !== undefined){
+                    $scope.order = data;
+                    // show the checkout page
+                    proccess_order_data();
+                    $scope.checkout_panel  = true;
+                }else if(data.crypto.code === 'btc'){
+                    $scope.address_error_btc = true;
+                }else if(data.crypto.code === 'bch'){
+                    $scope.address_error_bch = true;
+                }
```
shivaenigma, line 109: Server return empty data when new_address gen fails. So this will not work. Better to just use *crypto* variable that we already have.

js/app.js
```diff
@@ -131,392 +58,160 @@ app.controller('CheckoutController', function($scope, $interval, Order, $httpPar
...
-        }
-        return false;
-    }
-    function gotFunds(){
-        if(gotWrongAmount() || isPaymentReceived()){
-            return true;
+    //There are two styles of callback url in 
+    //woocommerce, we have to support both
+    //https://docs.woocommerce.com/document/wc_api-the-woocommerce-api-callback/
+    service.get_wc_endpoint = function(new_param = '', order_id = '', crypto_code = '') {
```
shivaenigma, line 194: This should be a generic function. It takes one dict() of parameter/value as input. Should not have specific parameters as given here (that limits the scope of this function a lot)

php/Blockonomics.php
```diff
@@ -238,4 +294,334 @@ public function testSetup()
...
+
+    public function check_for_late_payment($status, $order, $wc_order){
+        if($status == 0){
+            $time_period = get_option("blockonomics_timeperiod", 10) * 60;
+            if (time() > $order['timestamp'] + $time_period) {
+                $minutes = (time() - $order['timestamp']) / 60;
+                $wc_order->add_order_note(__("Warning: Payment arrived after $minutes minutes. Received". $order['crypto'] ."may not match current ". $order['crypto'] ." price", 'blockonomics-bitcoin-payments'));
+            }
+        }
+    }
```
shivaenigma, line 563: Remove this. Theortically this is useful, but practically it confuses merchants a lot. And we haven't see anyone using/requesting this

php/Blockonomics.php
```diff
@@ -238,4 +294,334 @@ public function testSetup()
...
+
+    // Save the new address to the WooCommerce order
+    public function record_address($order_id, $crypto, $address){
+        update_post_meta($order_id, $crypto .'_address', $address);
+    }
+
+    public function create_new_order($order_id, $crypto){
+        $responseObj = $this->new_address(get_option("blockonomics_callback_secret"), $crypto);
+        if($responseObj->response_code != 200) {
+            exit();
```
shivaenigma, line 441: It is a bad design to exit() without any return value atleast we should return an empty object. 
The calling JS is expecting the API to return something.

php/Blockonomics.php
```diff
@@ -238,4 +294,334 @@ public function testSetup()
...
+    public function check_for_pending_payment($orders, $order_id){
+        $network_confirmations = get_option("blockonomics_network_confirmation",2);
+        foreach ($orders as $address => $order){
+            if ($order['status'] >= 0 && $order['status'] < $network_confirmations){
+                $this->redirect_finish_order($show_order);
+            };
+        };
+    }
+
+    public function update_waiting_order($order){
```
shivaenigma, line 400: Rename to *calculate_order_params* as it is more appropriate

php/Blockonomics.php
```diff
@@ -238,4 +294,334 @@ public function testSetup()
...
+    // Fetch all the crypto orders linked to the order id
+    public function get_all_orders_by_id($orders, $order_id){
+        if(isset($orders[$order_id])){
+            return $orders[$order_id];
+        }
+        return false;
+    }
+
+    // Check and update the crypto order or create a new order
+    public function process_order($order_id, $crypto){
```
shivaenigma, line 496: Though the logic is right, it is confusing and a bit complex. Here is a simple logic (it doesn't cover some minor cases, but should be ok):
$order = get order with *order_id* and *crypto*
if ($order)
 $order = calculate_order_params($order)
else
  $order = create_new_order(order_id, crypto)

Notes:
*calculate_order_params* will not calculate if order status!=-1 or timestamp has not expired
For payment in process there is no extra logic required. When client gets these params along with correct timestamp, it will automatically redirect to order confirmation page after receiving tx from blockonomics

php/Blockonomics.php
```diff
@@ -238,4 +294,334 @@ public function testSetup()
...
+                $order = $this->create_new_order($order_id, $crypto);
+            }
+        }else {
+            $order = $this->create_new_order($order_id, $crypto);
+        }
+        
+        $orders[$order_id][$order['address']] = $order;
+        update_option('blockonomics_orders', $orders);
+
+        $order['crypto'] = $this->getActiveCurrencies()[$crypto];
```
shivaenigma, line 516: This looks pretty innocuous however, it causes lots of problem with object flow and structure.  In order object crypto is a 'string' everywhere. However, suddenly you combine two objects to return consolidate data to client. This distrurbs the object structures and creates problems 
For example, see in blockonomics in some app.JS we do multiple JS calls to fetch objects. We could theoretically combine all in single call to save time that we don't do that

A simple solution is not to have this and just render *active_cryptos* div in the crypto_checkout template also. Another solution is to add another JS endpoint for active cryptos

shivaenigma:
I am done with review. There is no need for noJS to be working. After these changes are done we will do testing and do a merge

blockonomics-woocommerce.php
DarrenWestwood: Fixed in f950572

js/app.js
DarrenWestwood: Fixed with 7e2bded

js/app.js
DarrenWestwood: Updated in 8686fdd

php/Blockonomics.php
DarrenWestwood: Removed in ac5506b

php/Blockonomics.php
DarrenWestwood: Updated in e8ef3ae

php/Blockonomics.php
DarrenWestwood: Renamed in 429b6c6

php/Blockonomics.php
DarrenWestwood: Updated in 69662b2

php/Blockonomics.php
DarrenWestwood: Fixed with 7e2bded and 7156d29

Issue #194, opened 2020-08-12
Lite mode display error is flashing while rendering payment screen


shivaenigma:
ng-cloak must be used

PR #198, opened 2020-08-21
Delay displaying lite-mode error

DarrenWestwood:
Fix issue #194 
Reinstates the class that that only shows the message after 3 seconds

### DIFF
--- a/templates/blockonomics_checkout.php
+++ b/templates/blockonomics_checkout.php
@@ -21,7 +21,7 @@ $blockonomics = new Blockonomics;
         <div class="bnomics-spinner"></div>
       </div>
       <!-- Display Error -->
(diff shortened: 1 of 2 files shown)


DarrenWestwood:
Lite mode errors are often due to other javascript errors breaking the execution of javascript on the checkout page. This causes our angular app to not run/remove the ng-cloak class. Using ng-cloak hides the lite-mode message, but does not reveal it if there is an error. 

In order for lite mode to function correctly the error is required to be displayed by default on the page and then get hidden when our javascript loads correctly. 
The _bnomics-display-error_ class delays the visibility of the lite-mode error to ensure the message does not display instantly, but gives some time for the angular app to load:
https://github.com/blockonomics/woocommerce-plugin/blob/master/css/order.css#L251

Issue #191, opened 2020-07-20
Payment expired early [server time issues]


shivaenigma:
*get_order* will return existing elapsed time from server in seconds. Client will then count forward from that

Time used in [websocket call](https://github.com/blockonomics/woocommerce-plugin/blob/396812db4650c3c71ce040ecba9213e373f2cad4/js/app.js#L238) is unnecessary and can be removed

Check any further places of timestamp use that maybe causing problems

DarrenWestwood:
### New Design


- The call to `get_order` should calculate the remaining time for the order and return the _remaining_time_ along with the order info instead of the _timestamp_.
- The client side should count/tick down from the _remaining_time_ to 0.
- Remove unnecessary passing of _blockonomics_time_period_ from the template and js: https://github.com/blockonomics/woocommerce-plugin/blob/191f029561be6d118ae9ef4286ca9bbf2c6175a7/templates/blockonomics_checkout.php#L122
- Remove the timestamp from the websocket. Verify all places the timestamp is used and check no problems are caused

shivaenigma:
Update the *timestamp* param from server shouldn't get removed. It is used to get the correct txid of payment payment (in case customer refersh the page at a very later time). 
We can send a new *remaining_time* parameter

PR #201, opened 2020-08-21
Fix payment expires early

DarrenWestwood:
Fix for issue #191

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -73,7 +73,7 @@ function CheckoutController($scope, $interval, Order, $timeout, Url) {
         }
         //Check the status of the order
         if ($scope.order.status === -1) {
-            $scope.clock = $scope.order.timestamp + totalTime - Math.floor(Date.now() / 1000);
+            $scope.clock = $scope.order.time_remaining;
             //Mark order as expired if we ran out of time
             if ($scope.clock < 0) {
                 $scope.order.status = -3;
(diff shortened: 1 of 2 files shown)


shivaenigma:
I have added back timestamp. It is required to handle cases when order is paid. In that case also server just returns order information. The timestamp is required to detect the tx that happened in past and then redirect to order confirmation page

shivaenigma:
On Page refresh it is always going back to 600 seconds. Always returning time remaining as 600 and timestamp is also changing(which shouldn't)

DarrenWestwood:
> I have added back timestamp. It is required to handle cases when order is paid. In that case also server just returns order information. The timestamp is required to detect the tx that happened in past and then redirect to order confirmation page

This was removed to try mitigate any undesired behaviour in the case of the an incorrect server timestamp being supplied. Using a backdated timestamp may lead to error messages such as `error_msg: "Invoice expired (Expiry time is 7 days). Please create a new invoice"`. From testing the websocket still seems to function as desired, even when set incorrectly.

> On Page refresh it is always going back to 600 seconds. Always returning time remaining as 600 and timestamp is also changing(which shouldn't)

I am unable to replicate this behaviour on chrome and firefox
Checkout refresh

The timestamp should only change if the existing order is expired. https://github.com/blockonomics/woocommerce-plugin/blob/master/php/Blockonomics.php#L408

PR #205, opened 2020-08-28
Fixed callback Issues

shivaenigma:
- order txid was not set before called record_payment method resulting in error. See screenshot
- simplified callback value compare logic
- In client, redirecting to order confirmation if status received indicates a successful transaction. Sometimes websocket may not work due to network disconnection or other issues. So UI wise it is preferable that refershing the page results in same behviour 
- Removed timestamp from websocket call to fix issues mentioned in https://github.com/blockonomics/woocommerce-plugin/pull/201

Screenshot from 2020-08-29 00-39-01

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -81,17 +81,23 @@ function CheckoutController($scope, $interval, Order, $timeout, Url) {
             }
             $scope.tick_interval = $interval($scope.tick, 1000);
             //Connect and Listen on websocket for payment notification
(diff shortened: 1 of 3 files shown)


DarrenWestwood (approved):
Callbacks are working well for exact amounts, underpayments and overpayments.
New logic looks good and is functioning correctly during tests.
Redirect when already pending payment is working.

PR #208, opened 2020-09-10
Add blockonomics version to enqueue scripts

DarrenWestwood:
Fix for lite mode issues when upgrading plugin versions.

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -361,16 +361,16 @@ function blockonomics_woocommerce_init()
     }
 
     function bnomics_enqueue_stylesheets(){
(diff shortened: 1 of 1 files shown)


DarrenWestwood:
When using _wp_enqueue_script_, If  the version parameter is false, a version number is automatically added equal to current installed WordPress version.
eg. app.js?ver=5.3.1
The Wordpress cache then still keeps using the old js until the wordpress version changes.

To fix, we now use the blockonomics version when loading the scripts.
eg. app.js?ver=2.0

With the changes in 80bd6a0 each Blockonomics version will clear the cache:
https://developer.wordpress.org/reference/functions/wp_enqueue_script/

Issue #209, opened 2020-09-11
Transaction info not added to order info


DarrenWestwood:
The plugin is no longer displaying the transaction link on customer invoices.
![Invoice](https://user-images.githubusercontent.com/39666372/92870952-b01da200-f404-11ea-8ca9-d2edadcd8d68.png)

This has not been updated for bch:
https://github.com/blockonomics/woocommerce-plugin/blob/master/blockonomics-woocommerce.php#L344-L361

shivaenigma:
The issue is that post meta is now in different format

   update_post_meta($wc_order->get_id(), 'paid_'. $order['crypto'], $value/1.0e8);

PR #211, opened 2020-09-11
Fix issue #209

DarrenWestwood:
Fixes issue #209 
Uses the new post meta titles to display the transaction info in the woocommerce order and email.

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -341,33 +341,34 @@ function blockonomics_woocommerce_init()
 
     <?php
     }
+    function bnomics_display_tx_info($order, $email=false)
+    {
+        include_once plugin_dir_path(__FILE__) . 'php' . DIRECTORY_SEPARATOR . 'Blockonomics.php';
+        $blockonomics = new Blockonomics();
+        $active_cryptos = $blockonomics->getActiveCurrencies();
+        foreach ($active_cryptos as $crypto) {
+            $txid = get_post_meta($order->get_id(), 'blockonomics_'.$crypto['code'].'_txid', true);
+            $address = get_post_meta($order->get_id(), $crypto['code'].'_address', true);
+            if ($txid && $address) {
+                if ($crypto['code'] == 'btc') {
+                    $base_url = Blockonomics::BASE_URL;
+                }else{
+                    $base_url = Blockonomics::BCH_BASE_URL;
+                }
+                echo '<h2>'.__('Payment Details', 'blockonomics-bitcoin-payments').'</h2><p><strong>'.__('Transaction', 'blockonomics-bitcoin-payments').':</strong>  <a href =\''. $base_url ."/api/tx?txid=$txid&addr=$address'>".substr($txid, 0, 10). '</a></p>';
+                if (!$email) {
+                   echo '<p>'.__('Your order will be processed on confirmation of above transaction by the bitcoin network.', 'blockonomics-bitcoin-payments').'</p>';
+                } 
+            }
+        }      
+    }
     function nolo_custom_field_display_cust_order_meta($order)
     {
-        $btc_txid = get_post_meta($order->get_id(), 'blockonomics_btc_txid', true);
-        $btc_address = get_post_meta($order->get_id(), 'btc_address', true);
-        include_once plugin_dir_path(__FILE__) . 'php' . DIRECTORY_SEPARATOR . 'Blockonomics.php';
-        if ($btc_txid && $btc_address) {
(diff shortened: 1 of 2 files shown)


blockonomics-woocommerce.php
```diff
@@ -343,20 +343,30 @@ function show_basic() {
...
         include_once plugin_dir_path(__FILE__) . 'php' . DIRECTORY_SEPARATOR . 'Blockonomics.php';
-        if ($txid && $address) {
-            echo '<h2>'.__('Payment Details', 'blockonomics-bitcoin-payments').'</h2><p><strong>'.__('Transaction', 'blockonomics-bitcoin-payments').':</strong>  <a href =\''. Blockonomics::BASE_URL ."/api/tx?txid=$txid&addr=$address'>".substr($txid, 0, 10). '</a></p><p>'.__('Your order will be processed on confirmation of above transaction by the bitcoin network.', 'blockonomics-bitcoin-payments').'</p>';
+        if ($btc_txid && $btc_address) {
+            echo '<h2>'.__('Payment Details', 'blockonomics-bitcoin-payments').'</h2><p><strong>'.__('Transaction', 'blockonomics-bitcoin-payments').':</strong>  <a href =\''. Blockonomics::BASE_URL ."/api/tx?txid=$btc_txid&addr=$btc_address'>".substr($btc_txid, 0, 10). '</a></p><p>'.__('Your order will be processed on confirmation of above transaction by the bitcoin network.', 'blockonomics-bitcoin-payments').'</p>';
+        }
+        $bch_txid = get_post_meta($order->get_id(), 'blockonomics_bch_txid', true);
+        $bch_address = get_post_meta($order->get_id(), 'bch_address', true);
```
shivaenigma, line 353: Code duplication and should be avoided. Use *txid* and *address* variable. *url* can also be a variable. It is unlikely that customer get both bch and btc tx in one order (this case can be ignored)

blockonomics-woocommerce.php
```diff
@@ -343,20 +343,30 @@ function show_basic() {
...
+        $btc_address = get_post_meta($order->get_id(), 'btc_address', true);
+        include_once plugin_dir_path(__FILE__) . 'php' . DIRECTORY_SEPARATOR . 'Blockonomics.php';
+        if ($btc_txid && $btc_address) {
+          echo '<h2>Payment Details</h2><p><strong>'.__('Transaction').':</strong>  <a href =\''. Blockonomics::BASE_URL ."/api/tx?txid=$btc_txid&addr=$btc_address'>".substr($btc_txid, 0, 10). '</a></p>';
+        }
+        $bch_txid = get_post_meta($order->get_id(), 'blockonomics_bch_txid', true);
+        $bch_address = get_post_meta($order->get_id(), 'bch_address', true);
+        if ($bch_txid && $bch_address) {
+          echo '<h2>Payment Details</h2><p><strong>'.__('Transaction').':</strong>  <a href =\''. Blockonomics::BCH_BASE_URL ."/api/tx?txid=$bch_txid&addr=$bch_address'>".substr($bch_txid, 0, 10). '</a></p>';
         }
```
shivaenigma, line 370: Avoid code duplication as above

DarrenWestwood:
Removes code duplication and fixes issue not displaying the transaction info before the transaction is confirmed.

php/Blockonomics.php
```diff
@@ -524,8 +525,7 @@ public function check_callback_secret($secret){
     }
 
     // Save the received payment info to the WooCommerce order
-    public function record_payment($value, $order, $wc_order){
-        update_post_meta($wc_order->get_id(), 'paid_'. $order['crypto'], $value/1.0e8);
+    public function record_transaction($value, $order, $wc_order){
```
shivaenigma, line 528: The code comment is still referring to the older. Both the names record_transaction adn record_payment are particularly similar, so we should attempt to clarify with correct code comments or more distinct function names

php/Blockonomics.php
```diff
@@ -538,6 +538,19 @@ public function is_payment_recorded($status, $order, $wc_order){
         return true;
     }
 
+    public function is_transaction_recorded($status, $order, $wc_order){
+        if (!metadata_exists('post',$wc_order->get_id(),'blockonomics_'. $order['crypto'] .'_txid', $order['txid']) )  {
+            return false;
+        }
+        return true;
+    }
```
shivaenigma, line 546: *is_transaction_recorded* and *record_transaction* can be merged into a single function. 
It is not a particularly good idea to keep having very small functions (unless they are used multiple times) as it leads to unnecessary function calls and code complexity

DarrenWestwood: Renamed and refactored in 9e0e8b2

PR #217, opened 2020-10-15
Fix nojs functionality

DarrenWestwood:
* Fixed nojs functionality
* Updated select order template to not use js
* All nojs functionality has been moved into a blockonomics_nojs_checkout template
* Updated nojs checkout ui to better match


shivaenigma:
Doing some preliminary testing. The BTC amount and address should be in a input box. That will help in copying (especially in mobile)

php/Blockonomics.php
```diff
@@ -297,11 +297,19 @@ public function testSetup()
     }
 
     // Create the order url to redirect the user to during checkout
```
shivaenigma, line 299: When I read the function comment, it was previously clear what the function was doing. Now with the *select* parameter it is completely confusing as the function comment and working doesn't match. I had to parse code for 10min to understand the logic. We need  functions to be modular - this means the input, output and working of the function should be understood just by reading the function itself. 

Here is a better structure:
```

get_order_checkout_url()
 // Returns url to redirect the user to during checkout
```

```
get_crypto_checkout_url($crypto, $order_id)
//Returns order url for checkout using a given cryptocurrency 
```

php/Blockonomics.php
```diff
@@ -450,13 +452,7 @@ public function load_checkout_template($order_id, $crypto){
...
     public function load_crypto_options_template(){
-        // Check to send the user to nojs page
-        if($this->is_nojs_active()){
-            $this->load_blockonomics_template('nojs_crypto_options');
-        }
-        else{
-            $this->load_blockonomics_template('crypto_options');
-        }
+        $this->load_blockonomics_template('crypto_options');
     }
```
shivaenigma, line 456: Quiz: Give two reasons why this function is not a good idea !

templates/blockonomics_nojs_checkout.php
```diff
@@ -1,70 +1,81 @@
...
+$qrcode = isset($_REQUEST['qrcode']) ? $_REQUEST['qrcode'] : "";
+
+if ($qrcode) {
+  $blockonomics->generate_qrcode($qrcode);
+}else if ($order_id) {
+  $order = $blockonomics->get_order_by_id_and_crypto($orders, $order_id, $crypto);
+  $order_url = $blockonomics->create_order_url($order_id, true);
+  if ($order['status'] >= 0){
+    $finish_order_url = add_query_arg('finish_order', $order['order_id'], $order_url);
+    wp_redirect($finish_order_url);
```
shivaenigma, line 15: This is unecessary, since you are already in php. This is there in app.js coz it is client side. Here you can directly call $blockonomics->redirect_finish_order()

templates/blockonomics_nojs_checkout.php
```diff
@@ -1,70 +1,81 @@
...
-}
-?>
-<div ng-app="shopping-cart-demo">
-  <div ng-controller="CheckoutController">
+$crypto = isset($_REQUEST["crypto"]) ? $_REQUEST["crypto"] : "";
+$order_id = isset($_REQUEST["show_order"]) ? $_REQUEST["show_order"] : "";
+$qrcode = isset($_REQUEST['qrcode']) ? $_REQUEST['qrcode'] : "";
+
+if ($qrcode) {
+  $blockonomics->generate_qrcode($qrcode);
```
shivaenigma, line 9: This code should be there in *WC_Gateway_Blockonomics.php/handle_requests()* 
That is where we are handling all the routing logic

php/Blockonomics.php
DarrenWestwood: Updated in 8e57418

php/Blockonomics.php
DarrenWestwood: Only used once and performs one function already covered. Removed in cbeaa28

templates/blockonomics_nojs_checkout.php
DarrenWestwood: Updated in 8496e81

templates/blockonomics_nojs_checkout.php
DarrenWestwood: Moved in f30f1d9

php/Blockonomics.php
shivaenigma: Yes it is unnecessarily creating an overhead of an extra function call and not providing and any modular improvement

php/Blockonomics.php
DarrenWestwood: Refactored code in 483b329

DarrenWestwood: Further refactored functions to be more usable in 412bc0b

Issue #222, opened 2020-11-30
Updating option should be atomic


blockonomics:
https://github.com/blockonomics/woocommerce-plugin/blob/master/php/Blockonomics.php#L488 use [blockonomics_order from line 479](https://github.com/blockonomics/woocommerce-plugin/blob/master/php/Blockonomics.php#L479). In between they are possible new_address calls , which will take 2-3 seconds. 

This may sometimes result in *blockonomics_orders* having stale values. There are callback threads also updating it. When order volume is a bit large, this may result in it updating *blockonomics_orders* with old/stale values

Here is the solution
- Read is fine and can continue like now 
- Just before write, a fresh read has to do done to fetch latest *blockonomics_order* from DB. We can move this to a function so that it is used everywhere

PR #224, opened 2020-12-21
Orders to db table

DarrenWestwood:
- Fixes issue #222 (atomic updates)
- Includes basic changes to move blockonomics_orders from wp options array to its own database table

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -437,11 +437,6 @@ function blockonomics_activation_hook() {
         require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
         dbDelta( $sql );
 
-        // Example function to demonstrate table changes between upgrade versions
-        // if ($installed_ver < 1.0) {
-        //     $wpdb->query("ALTER TABLE $table_name DROP transaction;");
-        // }
-
         update_option( 'blockonomics_db_version', $blockonomics_db_version );
     }
 
@@ -451,8 +446,16 @@ function blockonomics_activation_hook() {
 // Since WP 3.1 the activation function registered with register_activation_hook() is not called when a plugin is updated.
 function blockonomics_update_db_check() {
     global $blockonomics_db_version;
-    if ( get_site_option( 'blockonomics_db_version' ) != $blockonomics_db_version ) {
-        blockonomics_activation_hook();
+
+    $installed_ver = get_site_option( 'blockonomics_db_version' );
+    if ( $installed_ver != $blockonomics_db_version ) {
+    	
+        // Example function to demonstrate table changes between upgrade versions
+        // if ($installed_ver < 1.0) {
+        //     $wpdb->query("ALTER TABLE $table_name DROP transaction;");
+        // }
+
+        update_option( 'blockonomics_db_version', $blockonomics_db_version );
     }
 }
 add_action( 'plugins_loaded', 'blockonomics_update_db_check' );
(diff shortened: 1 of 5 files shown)


DarrenWestwood:
Moved blockonomics_orders from wp options array to its own database table:
https://codex.wordpress.org/Creating_Tables_with_Plugins

Added DB version for upgrade procedure:
https://codex.wordpress.org/Creating_Tables_with_Plugins#Adding_an_Upgrade_Function

Added sanitization to request data:
https://codex.wordpress.org/Validating_Sanitizing_and_Escaping_User_Data#Sanitizing:_Cleaning_User_Input

Correctly prepares $wpdb->get_row() sql statements:
https://codex.wordpress.org/Data_Validation#Database
This is [not required](https://wordpress.stackexchange.com/q/51057) for $wpdb->replace()

jusasiiv (approved):
Code is good, did basic functional testing with upgrading plugin in existing installation

jusasiiv (commented):
WC_Gateway_Blockonomics line 30 and ...nojs_checkout.php line 3 have reference to blockonomics_orders option which is no longer used

blockonomics-woocommerce.php
```diff
@@ -401,14 +401,62 @@ function bnomics_async_scripts($url)
...
+            status int NOT NULL,
+            crypto varchar(3) NOT NULL,
+            address varchar(191) NOT NULL,
+            timestamp int,
+            time_remaining int,
+            satoshi int,
+            currency varchar(3),
+            value longtext,
+            txid text,
+            UNIQUE (address),
```
blockonomics, line 434: Unique keyword is unnecessary. We already are using PRIMARY KEY

blockonomics-woocommerce.php
```diff
@@ -401,14 +401,62 @@ function bnomics_async_scripts($url)
...
 }
 
+// Since WP 3.1 the activation function registered with register_activation_hook() is not called when a plugin is updated.
+function blockonomics_update_db_check() {
+    global $blockonomics_db_version;
+    if ( get_site_option( 'blockonomics_db_version' ) != $blockonomics_db_version ) {
+        blockonomics_activation_hook();
+    }
+}
+add_action( 'plugins_loaded', 'blockonomics_update_db_check' );
```
blockonomics, line 458: Since we are anyway using the *plugins_loaded* hook, I think it is uncessary to add the update code inside activation hook. This will call it to be invoked twice - once on activation and another on *plugins_loaded* . We can move this code to a seperate function

blockonomics:
Some testing done. Further testing waiting for https://github.com/blockonomics/woocommerce-plugin/issues/228 to be resolved.
Also lets remove the code `delete_option('blockonomics_orders');`

blockonomics:
- Also delete plugin should remove our table (lets use WP recommended way on this)

blockonomics-woocommerce.php
DarrenWestwood: Updated in 38893c1 and fa71700

blockonomics-woocommerce.php
DarrenWestwood: WordPress recommends the upgrade be structured in this manner. https://codex.wordpress.org/Creating_Tables_with_Plugins#Adding_an_Upgrade_Function

I have separated the install and upgrade function in ef135d6. Users upgrading from a previous version will now be required to deactivate and activate the plugin in order to create the table.

DarrenWestwood: Fixed logic for existing users in c79436f

DarrenWestwood:
> * Also delete plugin should remove our table (lets use WP recommended way on this)
Updated in 8aeb424

PR #232, opened 2021-02-10
Wc filter

cnohall:
It works, please let me know what you think and how it might be improved.


php/WC_Filter.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+	public function filter_orders_by_address() {
+		global $typenow;
+
+		// add_settings_error('option_notice', 'option_notice', 'Hello', 'error');
+		if ( 'shop_order' === $typenow ) {
+			?>
+			<input type='name' placeholder='Filter bitcoin payments' name='filter_by_address'>
+			<select name="type_query" id="dropdown_type_query">
+				<option value="btc_address">
```
DarrenWestwood, line 22: Filtering by the btc_address, btc_txid, and bch_txid is working correctly. Filter by bch_address should also be added.

DarrenWestwood (changes requested):

DarrenWestwood:
Would it be possible to combine the filters into a single input?
Orders-‹-Wordpress-—-WordPress (1)

There is no need to implement this yet.

cnohall:
It should be possible. I tried finding a solution, but after a while I went
for two fields because it was easier to implement. I'll continue to look
into a more elegant solution.

php/WC_Filter.php
cnohall: added bch_address as an option.

cnohall (commented):
I'll continue to look into how to merge the two fields into one.

blockonomics:
Just a UI thing - In UI we don't need to mention filter by BCH address/txid or filter by BTC address/txid. We should only say *Filter by crypto address/txid* 

If it by default all custom crypto fields in an order

cnohall:
Understood, I think I managed to shrink it to one field that is capable of
both btc/bch and address/txid. I've pushed the changes. Give it a try and
let me know what you think.

DarrenWestwood:
The changes are neat and function as planned.

php/WC_Filter.php
```diff
@@ -0,0 +1,41 @@
...
+			add_action( 'restrict_manage_posts', array( $this, 'filter_orders') , 20 );
+			add_filter( 'request',               array( $this, 'filter_orders_by_address_or_txid' ) );	
+
+		}
+	}
+	public function filter_orders() {
+		global $typenow;
+		if ( 'shop_order' === $typenow ) {
+			?>
+			<input size='26' type='name' placeholder='Filter by crypto address/txid' name='filter_by'>
```
DarrenWestwood, line 19: When the filter is applied, the page refreshes and clears the address/txid from the input. We can use `value="<?php if(isset( $_GET['filter_by'] )) echo($_GET['filter_by']); ?>"` to set the value of the input to display the filtered address/txid.

cnohall: Great idea! I've added it.

DarrenWestwood (approved):

shivaenigma:
Functionality looks good. I am just concerned by the use of *plugins_loaded* action. As far as I tested (you can verify), it is called on every wordpress page visited so I am not sure this is a correct place to add - maybe find some woocommerce related actions.

cnohall:
I'll look into the _plugins_loaded_ hook/action and see how it might be improved. I'll also examine the WooCommerce plugin closer and see what hooks they use, and opt to go for best practice.

cnohall:
I did some testing (using S*how Hooks *by rafiq91). You're correct, the
wc-filter is called on every wordpress page visited.

I changed the code to only activate the filter on the WooCommerce->Order
page through an if check:

if(isset($_GET['post_type']) && $_GET['post_type'] == 'shop_order') {
    add_action( 'plugins_loaded', 'wc_filter_orders_by_bitcoin_address' );
} else {
    remove_filter( 'plugins_loaded', 'wc_filter_orders_by_bitcoin_address'
 );
}

After adding this check and doing some testing the wc-filter appears to
only get loaded when the user/admin visits the WooCommerce->Order page.

cnohall:
Removed reliance on "plugins_loaded", now it only relies on "restrict_manage_posts" and "requests".

cnohall:
Removed need for the file wc_filter.php.  Now the code is less than 20 lines long and should be easy to follow and provide the wished for result. The only action needed is '[restrict_manage_posts](https://woocommerce.github.io/code-reference/classes/WC-Admin-List-Table.html#method_restrict_manage_posts)' provided by WooCommerce.

PR #234, opened 2021-02-16
UI refactoring

cnohall:
The UI refactoring should be in a working condition. Let me know if it can be improved in any way.


blockonomics-woocommerce.php
```diff
@@ -223,122 +218,232 @@ function show_basic() {
...
+                    break;
+
+
+
+
+                case 'currencies' :
+                    ?>
+                    <table style="margin-left:40px; padding: 0px 0px !important; " class="form-table">
+                    <h1>
+                        <input style="margin-left:12px;" type="checkbox" name="blockonomics_btc" value="1"<?php checked("1", get_option('blockonomics_btc')); ?>" />
```
DarrenWestwood, line 310: Bitcoin should be enabled by default if not set yet- `get_option("blockonomics_btc", true);`

blockonomics-woocommerce.php
```diff
@@ -223,122 +218,232 @@ function show_basic() {
...
-                            </p>
-                            <?php else: ?>
-                            <p> To withdraw, Click on <b>Test Setup</b></p>
+                <?php wp_nonce_field('update-options');
+                switch ( $active_tab ){
+                case 'settings' :
+                    ?>
+                    
+                    <h1>API Key</h1>
+                    <input onchange="value_changed()" size="130" type="text" id="blockonomics_api_key" name="blockonomics_api_key" value="<?php echo get_option('blockonomics_api_key'); ?>" />
```
DarrenWestwood, line 240: The inputs are not responsive on smaller screens. Add `max-width: 100%;` to fix and `text-overflow: ellipsis;` to indicate if text is longer than the input.

blockonomics-woocommerce.php
```diff
@@ -223,122 +218,232 @@ function show_basic() {
...
+                                <td><select name="blockonomics_network_confirmation" />
+                                        <option value="2" <?php selected(get_option('blockonomics_network_confirmation'), 2); ?>>2 (Recommended)</option>
+                                        <option value="1" <?php selected(get_option('blockonomics_network_confirmation'), 1); ?>>1</option>
+                                        <option value="0" <?php selected(get_option('blockonomics_network_confirmation'), 0); ?>>0</option>
+                                    </select></td>
+                            </tr>
+                        </table>
+                    </div>
+                    <p class="submit">
+                        <input type="submit" class="button-primary" value="Save"/>
```
DarrenWestwood, line 290: The Save and Test Setup buttons do not match. Add `padding: 3px 15px 3px 15px;` and `font-size: 15px;` to the Save button.

blockonomics-woocommerce.php
```diff
@@ -223,122 +218,232 @@ function show_basic() {
...
+                </form>
+                    <?php
+                    break;
+
+
+
+
+                case 'currencies' :
+                    ?>
+                    <table style="margin-left:40px; padding: 0px 0px !important; " class="form-table">
```
DarrenWestwood, line 308: Some items in the currencies tab shift off the screen to the right on smaller devices.
Small screen

DarrenWestwood (changes requested):

blockonomics-woocommerce.php
```diff
@@ -109,26 +107,18 @@ function add_page()
...
-            {
-                $message = __('Error while generating temporary APIKey: '. isset($response->message) ? $response->message : '', 'blockonomics-bitcoin-payments');
-                display_admin_message($message, 'error');
-            }
-            else
-            {
-                update_option("blockonomics_temp_api_key", isset($response->apikey) ? $response->apikey : '');
-            }
+            $btc_api_key_response = $blockonomics->get_temp_api_key($callback_url, 'btc');
+            $bch_api_key_response = $blockonomics->get_temp_api_key($callback_url, 'bch');
```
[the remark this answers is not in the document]
cnohall: Removed it.

php/Blockonomics.php
```diff
@@ -217,85 +337,56 @@ private function set_headers($api_key)
...
+    // Runs when the Blockonomics Test Setup button is clicked
+    // Returns any errors or false if no errors
+    public function testSetup()
+    {
+        $test_results['btc'] = false;
+        $test_results['bch'] = false;
+        // Fetch the crypto to test based on the plugin settings
+        $enabled_cryptos = $this->get_test_setup_crypto();
+        if ($enabled_cryptos['btc_enabled']){
+            $test_results['btc'] = $this->test_one_crypto('btc');
```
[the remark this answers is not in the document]
cnohall: Awesome, I've simplified testSetup() according to the code above.

blockonomics-woocommerce.php
cnohall: Made and committed the change.

blockonomics-woocommerce.php
cnohall: Done, made the input fields responsive.

blockonomics-woocommerce.php
```diff
@@ -142,27 +132,38 @@ function add_page()
...
 
-        if (isset($_POST['runTest']))
+
+        if (isset($_GET['tab']) && $_GET['tab'] == "currencies" && isset($_GET['settings-updated']) ? $_GET['settings-updated'] : '' == 'true')
         {
             $setup_errors = $blockonomics->testSetup();
-
-            if($setup_errors)
+            update_option("setup_errors", $setup_errors);
+            if(!$setup_errors['bch'] && !$setup_errors['btc'])
```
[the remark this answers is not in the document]
cnohall: Done.

blockonomics-woocommerce.php
cnohall: Done.

php/Blockonomics.php
```diff
@@ -217,85 +337,56 @@ private function set_headers($api_key)
...
 
-    public function testSetup()
-    {
-        $response = $this->get_callbacks();
-        $error_str = '';
-        $response_body = json_decode(wp_remote_retrieve_body($response));
-        if(isset($response_body[0])){
-            $response_callback = isset($response_body[0]->callback) ? $response_body[0]->callback : '';
-            $response_address = isset($response_body[0]->address) ? $response_body[0]->address : '';
+    public function get_test_setup_crypto() {
```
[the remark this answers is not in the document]
cnohall: Updated get_order_checkout_url() and getActiveCurrencies() to handle the case if only bch is enabled.

shivaenigma:
We should allow merchant to disble both BTC and BCH currencies. That is fine. On checkout the the crypto selection screen should say

*No crypto currencies are enabled for checkout* 
Note to webmaster: Can be enabled via Wordpress Admin > Settings > Blockonomics > Currencies

cnohall:
[This ](https://drive.google.com/file/d/1ak_u8W6pyOCnax1YpblUQvNoee5SXicc/view?usp=sharing) is how the checkout screen looks like now when no cryptos are active. Please let me know if the design can be improved in any way.

blockonomics-woocommerce.php
```diff
@@ -223,122 +218,232 @@ function show_basic() {
...
+                case 'settings' :
+                    ?>
+                    
+                    <h1>API Key</h1>
+                    <input onchange="value_changed()" size="130" type="text" id="blockonomics_api_key" name="blockonomics_api_key" value="<?php echo get_option('blockonomics_api_key'); ?>" />
+                    <p>To get your API Key, click <b> Get Started for Free </b> on
+                        <a href="https://blockonomics.co.merchants">https://blockonomics.co.merchants</a>
+                    </p>
+                    <h1>Callback URL
+                        <a href="javascript:gen_secret()" id="generate-callback" style="font:400 30px/1 dashicons;margin-left: 5px;top: 6px;position:relative;text-decoration: none;" title="Generate New Callback URL">&#xf463;</a>
```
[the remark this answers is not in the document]
cnohall: I've now created a file called `blockonomics_options.css` for the setting page CSS and made sure that it only loads on the `blockonomics_options `page using `admin_enqueue_scripts `. I've also migrated the style code from the options page into `blockonomics_options.css`.

blockonomics-woocommerce.php
cnohall: I've corrected it.

cnohall:
Both BTC and BCH work separately, and together. If none of the cryptos are chosen the user will be shown a message at the checkout screen. The styles have been migrated from the `blockonomics-woocommerce.php` to `blockonomics_options.css` and made responsive.

cnohall (commented):
I've done the changes according to Darren's comments

cnohall (commented):
Done

blockonomics-woocommerce.php
```diff
@@ -60,11 +60,24 @@ function blockonomics_woocommerce_init()
...
     add_filter('woocommerce_payment_gateways', 'woocommerce_add_blockonomics_gateway');
     add_filter('clean_url', 'bnomics_async_scripts', 11, 1 );
 
+
+    /**
+     * Add Styles to Blockonomics Admin Page
+     **/
+    function blockonomics_load_admin_scripts(){ 
+        $current_screen = get_current_screen();
+        if ( strpos($current_screen->base, 'settings_page_blockonomics_options') === false) {
```
DarrenWestwood, line 75: admin_enqueue_scripts has a $hook parameter that can be used. `if ( $hook === 'settings_page_blockonomics_options' )`

blockonomics-woocommerce.php
```diff
@@ -243,121 +240,228 @@ function show_basic() {
...
+                        <table class="form-table bnomics-options-intendation-heading">
+                        <?php 
+                        $bch_enabled = get_option("blockonomics_bch");
+                        if ($bch_enabled):  ?>
+                        <th scope="row"><h1>Destination</h1></th>
+                                <td colspan="2" class="bnomics-options-no-padding">
+                                    <?php
+                                    $total_received = get_option('blockonomics_temp_withdraw_amount') / 1.0e8;
+                                    $api_key = get_option("blockonomics_api_key");
+                                    $temp_api_key = get_option("blockonomics_temp_api_key");
```
DarrenWestwood, line 407: Temp wallet functionality can be removed from BCH.

php/Blockonomics.php
```diff
@@ -309,17 +365,20 @@ public function get_parameterized_wc_url($params = array()){
...
     public function get_order_checkout_url($order_id){
+        $active_cryptos = $this->getActiveCurrencies();
         // Check if more than one crypto is activated
-        if (count($this->getActiveCurrencies()) > 1) {
+        if (count($active_cryptos) > 1) {
             $order_url = $this->get_parameterized_wc_url(array('select_crypto'=>$order_id));
-        }
-        // Default to btc if only bitcoin is active
-        else{
+        } elseif (isset($active_cryptos['btc']['code']) ){
```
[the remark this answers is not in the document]
cnohall: For some reason I couldn't get it to work with `Object.keys`. It seems like it's a JavaScript function? I get the wished for result by using `'crypto'=>reset(reset($active_cryptos))`. Perhaps there is an even better solution? In the meantime I'll push the changes.

blockonomics-woocommerce.php
cnohall: Awesome, that makes a lot of sense and is a much better solution. Thanks! I've pushed the changes

blockonomics-woocommerce.php
cnohall: True, that also makes the code more concise and easier to work with. I've made the changes and pushed them.

cnohall:
I've made all the changes. However, I'm not certain if `'crypto'=>reset(reset($active_cryptos))` is acceptable or if there is a more suitable solution.

DarrenWestwood:
> I've made all the changes. However, I'm not certain if `'crypto'=>reset(reset($active_cryptos))` is acceptable or if there is a more suitable solution.
It may be better to try using Object.keys($active_cryptos)[0] or $active_cryptos[0]['code'] as this may be a cleaner solution.

DarrenWestwood:
> I've made all the changes. However, I'm not certain if `'crypto'=>reset(reset($active_cryptos))` is acceptable or if there is a more suitable solution.

It may be better to try array_key_first(), as this may be a cleaner solution

cnohall:
Nice, array_key_first() did the job!

blockonomics:
Just tried few tests.

- Why am I getting https://prnt.sc/107erfb everytime I click on Test Setup ? We should always think about customer experience
- BCH test setup code has some bug. I am getting error even though BCH checkout is working for me

blockonomics:
- When user changes tab --lets trigger a auto save on tab only if the changes are not already saved (no asterix present)
-  No need for *Settings Saved* message at all in both tabs- this is annoying and can be completely removed
- Asterix thing on unsaved tab is looking good.  Would be good to have a [subtle color on the tab too](https://ux.stackexchange.com/questions/23137/denoting-a-tab-as-being-unsaved-in-a-web-application)

cnohall:
The _Settings Saved_ have now been removed from both tabs. I'm moving on to auto saving when changing between the two tabs in the plugin. 

Do we want to disable any cryptos that come back with an error when running the setup test? It makes sense to not save a crypto as enabled, if it's not set up right. However, I'm eager to hear your thoughts on it.

blockonomics:
>Do we want to disable any cryptos that come back with an error when running the setup test? It makes sense to not save a crypto as enabled, if it's not set up right. However, I'm eager to hear your thoughts on it.

No lets not disable any crypto automatically. User will do it if it he wants.

cnohall:
- When user changes tab --lets trigger an auto save on tab only if the changes are not already saved (no asterix present).
- No need for Settings Saved message at all in both tabs- this is annoying and can be completely removed
- Asterix thing on unsaved tab is looking good. Would be good to have a subtle color on the tab too.

These are now fininshed. However, once a user has made changes to the form it is registered as _changed_. This means that the autosave, the Asterix etc remain, even if the user goes back to an _unchanged_ state. This might be confusing to the user. 

I'll work on finding a way of returning the state to _unchanged_ if the user reverts the form to its original state.

blockonomics:
>These are now fininshed. However, once a user has made changes to the form it is registered as changed. This means that the autosave, the Asterix etc remain, even if the user goes back to an unchanged state

This is uncessary and will create unneeded complexity. Even editors like Wordpres etc don't do this

cnohall (commented):
Changes discussed have been made. It should now be ready for a new review.

css/blockonomics_options.css
```diff
@@ -0,0 +1,125 @@
...
+    cursor: default !important;
+}
+
+.wp-core-ui .notice.is-dismissible {
+    display: none;
+}
+
+
+.form-table th {
+    width: 90px !important;
```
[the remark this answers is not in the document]
cnohall: Created a new class _.blockonomics-narrow-th_ so that it only affects the relevant elements.

blockonomics-woocommerce.php
```diff
@@ -242,122 +276,197 @@ function show_basic() {
...
-                            </p>
-                            <?php else: ?>
-                            <p> To withdraw, Click on <b>Test Setup</b></p>
+                <input type="hidden" name="blockonomics_redirect" id="blockonomics_redirect" value="false">
+                <?php wp_nonce_field('update-options');
+                switch ( $active_tab ){
+                case 'settings' :?>
+                    <h4>API Key</h4>
+                    <input class="bnomics-options-input" onchange="value_changed()" size="130" type="text" id="blockonomics_api_key" name="blockonomics_api_key" value="<?php echo get_option('blockonomics_api_key'); ?>" />
+                    <label><br/>To get your API Key, click <b> Get Started for Free </b> on
```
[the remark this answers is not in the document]
cnohall: Removed the `br` tag and removed the padding for the buttons.

css/blockonomics_options.css
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.form-table td p {
+    margin-top: 0px !important;
+}
+
+.form-table {
+    width: 50vh !important;
```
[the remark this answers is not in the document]
cnohall: Done.

blockonomics-woocommerce.php
```diff
@@ -155,26 +162,28 @@ function add_page()
             'Blockonomics', 'Blockonomics', 'manage_options',
             'blockonomics_options', 'show_options'
         );
-
-        if (get_option('blockonomics_api_updated') == 'true' && isset($_GET['settings-updated']) ? $_GET['settings-updated'] : '' == 'true')
+        
+        if (get_option('blockonomics_redirect') == "yes")
```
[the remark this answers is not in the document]
cnohall: The `blockonomics_redirect` is as saved as part of the form so that we know if we should redirect the user after submitting the form. 

If a user has made changes to the form and then clicks a different tab within the plugin the form will be submitted and `blockonomics_redirect` will be saved as 'yes'. After the form is submitted a check will be done and if the user should be redirected `blockonomics_redirect` will be set back to 'no' and the redirect take place.

blockonomics-woocommerce.php
```diff
@@ -60,11 +60,21 @@ function blockonomics_woocommerce_init()
     add_action('init', 'load_plugin_translations');
     add_action('woocommerce_order_details_after_order_table', 'nolo_custom_field_display_cust_order_meta', 10, 1);
     add_action('woocommerce_email_customer_details', 'nolo_bnomics_woocommerce_email_customer_details', 10, 1);
-    add_action( 'restrict_manage_posts', 'filter_orders' , 20 );
-    add_filter( 'request', 'filter_orders_by_address_or_txid' );	
+    add_action('admin_enqueue_scripts', 'blockonomics_load_admin_scripts' );
+    add_action('restrict_manage_posts', 'filter_orders' , 20 );
+    add_filter('request', 'filter_orders_by_address_or_txid' );
```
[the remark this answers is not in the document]
cnohall: removed it.

blockonomics-woocommerce.php
```diff
@@ -155,26 +162,28 @@ function add_page()
...
+        if (isset($_GET['tab']) && $_GET['tab'] == "currencies" && isset($_GET['settings-updated']) ? $_GET['settings-updated'] : '' == 'true')
         {
             $setup_errors = $blockonomics->testSetup();
-
-            if($setup_errors)
-            {
-                display_admin_message($setup_errors, 'error');
+            set_transient('setup_test_finished', true, 1);
+            if(isset($setup_errors['bch'])){
+                set_transient('setup_error_bch', $setup_errors['bch'], 1);
```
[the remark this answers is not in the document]
cnohall: I moved the result from the test setup to the URL as params. This makes it possible to use the options api or the transient api.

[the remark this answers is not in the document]
DarrenWestwood: There is no need to pass the error message through the URL parameter. By moving the test setup into the `show_options()` function, the errors will be directly accessible in `$setup_errors`
```
function show_options()
    {
    	if (isset($_GET['tab']) && $_GET['tab'] == "currencies" && isset($_GET['settings-updated']) ? $_GET['settings-updated'] : '' == 'true')
        {
            $setup_errors = $blockonomics->testSetup();
```

DarrenWestwood (commented):
The UI is looking good and matches the new design. There are still some unnecessary margins and padding on the save button.

blockonomics-woocommerce.php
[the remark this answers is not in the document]
cnohall: Now I see, thanks for the help! I've made the changes and pushed it.

cnohall:
The margins and paddings seems to be set from wp and not us. I removed the class that used to be associated with the _save_ and _test setup_ buttons. Hopefully that will make it look like we want to.

blockonomics-woocommerce.php
```diff
@@ -496,7 +586,7 @@ function blockonomics_plugin_activation() {
 
     $html = '<div class="updated">';
     $html .= '<p>';
-    $html .= __( 'Congrats, you are now accepting BTC payments! You can configure Blockonomics <a href="options-general.php?page=blockonomics_options">on this page</a>.', 'blockonomics-bitcoin-payments' );
+    $html .= __( 'Congrats, you are now accepting BTC payments! You can configure Blockonomics <a href="page=options-general.php?blockonomics_options">on this page</a>.', 'blockonomics-bitcoin-payments' );
```
DarrenWestwood, line 589: This URL leads to the incorrect page after plugin activation.

blockonomics-woocommerce.php
```diff
@@ -155,30 +142,16 @@ function add_page()
...
-                $message = __('Congrats ! Setup is all done', 'blockonomics-bitcoin-payments');
-                display_admin_message($message, 'updated');
-
-                $message = $blockonomics->make_withdraw();
-                if ($message) {
-                    display_admin_message($message[0], $message[1]);
-                }
+            update_option('blockonomics_redirect', 'no');
+            if(isset($_GET['tab']) && $_GET['tab'] == "currencies"){
+                wp_redirect( home_url('/wp-admin/options-general.php?page=blockonomics_options&tab=settings') );
```
DarrenWestwood, line 150: The admin settings are functioning well and the UI is complete. One last improvement would be to use `admin_url( 'options-general.php?page=blockonomics_options&tab=settings
' );` instead of `home_url()` to form the redirection URL.

DarrenWestwood, line 150: The admin settings are functioning well and the UI is complete. One last improvement would be to use `admin_url( 'options-general.php?page=blockonomics_options&tab=settings
' );` instead of `home_url()` to form the redirection URL.

blockonomics:
Following are results of testing:

- If I select one crypto and do checkout, I get the following error:
`[10-Mar-2021 13:45:28 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function array_key_first() in /home/enigma/blockonomics-bitcoin-payments/php/Blockonomics.php:371`
- I don't like the ellipses thing https://prnt.sc/10i1vuy. The user should be able to navigate right in the text bx and see the complete URL or if that is not possible lets go to it being a label as before
- Temp Wallet configuration issues . To test try plugin deactivate, delete and then activate again
   - Lets format and correct instructions https://prnt.sc/10i1zte
   - Correct instructions https://prnt.sc/10i237j
 We are getting closer to release !

blockonomics-woocommerce.php
cnohall: Makes sense. Thanks for the tip. I've committed and pushed the changes.

blockonomics-woocommerce.php
cnohall: It's now been updated.

blockonomics (approved):

shivaenigma:
Almost ready for release. Few bugs to be fixed:
- Go into temp wallet mode, deactive > Delete > activate plugin. Do checkout (without going into Settings). It says no currencies. This is untrue as we do have active BTC. Once we click on Test Setup, it starts working
- All strings should be localized like this
```
<?php echo __('Network Confirmations required for payment to complete)', 'blockonomics-bitcoin-p
ayments')?>
```
- Why are we [getting clickable mouse icon over our text](https://drive.google.com/file/d/16SrnjBQUSDKTKXaXZV8cFDbepYo-1jCW/view?usp=sharing) that is non clickable ! It should behave like normal Text

shivaenigma:
Functionality looks ok to me for approval. @DarrenWestwood if you can have a final look at the code we can then proceed with a merge

css/blockonomics_options.css
```diff
@@ -0,0 +1,133 @@
...
+
+.bnomics-options-red-cross-mark {
+    font:1000 25px/1 dashicons; 
+    top:0px; 
+    color:red
+}
+
+
+.bnomics-options-small-intendation {
+    margin-left:12px !important;
```
DarrenWestwood, line 104: The Settings Tab and Currencies Tab have different left margins, which appear to be unnecessary.

blockonomics-woocommerce.php
```diff
@@ -60,29 +60,19 @@ function blockonomics_woocommerce_init()
...
+    add_action('admin_enqueue_scripts', 'blockonomics_load_admin_scripts' );
     add_filter('woocommerce_payment_gateways', 'woocommerce_add_blockonomics_gateway');
     add_filter('clean_url', 'bnomics_async_scripts', 11, 1 );
 
+
     /**
-     * Adding new filter to WooCommerce orders
+     * Add Styles to Blockonomics Admin Page
      **/
-    function filter_orders() {
```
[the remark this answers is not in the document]
shivaenigma: It already has merged *origin/master* . Github will not bring in such changes by itself. This is a mistake during conflict resolution

@cnohall  please be careful with conflict resolution. This is very critical. This will potentially create/remove major sections of code after are unwanted. You may want to read more about using git

php/Blockonomics.php
```diff
@@ -16,10 +16,21 @@ class Blockonomics
...
     const BCH_PRICE_URL = 'https://bch.blockonomics.co/api/price';
+    const BCH_SET_CALLBACK_URL = 'https://bch.blockonomics.co/api/update_callback';
+    const BCH_GET_CALLBACKS_URL = 'https://bch.blockonomics.co/api/address?&no_balance=true&only_xpub=true&get_callback=true';
+
 
     public function __construct()
     {
         $this->api_key = $this->get_api_key();
+        //array_key_first was introduced in PHP 7.3. This is to keep backwards compability
+        if (!function_exists('array_key_first')) {
```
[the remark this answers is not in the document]
cnohall: Smart! I've done the changed and pushed it.

css/blockonomics_options.css
cnohall: I removed much of the left margins in the Currencies Tab. Let me know what you think.

blockonomics-woocommerce.php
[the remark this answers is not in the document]
cnohall: I've rebased the PR. I'll be more careful in conflict resolution and will study git more closely so that it doesn't happen again.

cnohall:
Hopefully I did the merging right. When I tried to rebase it wanted me to manually resolving over 100 conflicts. Not sure that that's the right way.

php/Blockonomics.php
```diff
@@ -79,23 +92,119 @@ public function get_price($currency, $crypto)
...
+            $response_address = '';
+            if(isset($response_body[0])){
+                $response_callback = isset($response_body[0]->callback) ? $response_body[0]->callback : '';
+                $response_address = isset($response_body[0]->address) ? $response_body[0]->address : '';
+            }
+            $response_callback_without_schema = preg_replace('/https?:\/\//', '', $response_callback);
+            if(!$response_callback || $response_callback == null)
+            {
+                //No callback URL set, set one 
+                $this->update_callback($callback_url, $response_address, $crypto);
```
DarrenWestwood, line 156: update_callback should receive `$callback_url, $crypto, $xpub`, however we are passing `$callback_url, $response_address, $crypto`. This leads to "Could not find matching xpub." error.

cnohall: Fixed

blockonomics-woocommerce.php
```diff
@@ -242,123 +267,176 @@ function show_basic() {
...
-                            <p> To withdraw, Click on <b>Test Setup</b></p>
+                <input type="hidden" name="blockonomics_redirect" id="blockonomics_redirect" value="false">
+                <?php wp_nonce_field('update-options');
+                switch ( $active_tab ){
+                case 'settings' :?>
+                <div class="bnomics-width">
+                    <h4><?php echo __('API Key', 'blockonomics-bitcoin-payments')?></h4>
+                    <input class="bnomics-options-input" onchange="value_changed()" size="130" type="text" id="blockonomics_api_key" name="blockonomics_api_key" value="<?php echo get_option('blockonomics_api_key'); ?>" />
+                    <p class='bnomics-options-no-margin'><?php echo __('To get your API Key, click <b> Get Started for Free </b> on', 'blockonomics-bitcoin-payments')?>
+                        <a href="https://blockonomics.co.merchants"><?php echo __('https://blockonomics.co.merchants', 'blockonomics-bitcoin-payments')?></a>
```
DarrenWestwood, line 303: The links to Blockonomics should open in a new browser tab. Add `target="_blank"` to the Get started links.

cnohall: Done.

shivaenigma:
>The blockonomics_redirect is as saved as part of the form so that we know if we should redirect the user after submitting the form.

If a user has made changes to the form and then clicks a different tab within the plugin the form will be submitted and blockonomics_redirect will be saved as 'yes'. After the form is submitted a check will be done and if the user should be redirected blockonomics_redirect will be set back to 'no' and the redirect take place.

This should be possible without using an option. Lets think about this and design an elegant solution

cnohall:
So far we have tried: 

- using the URL and adding a param that can be read after the form has been submitted.
- [Transients](https://developer.wordpress.org/apis/handbook/transients/). 

I'm currently looking into using [admin_post_$action](https://developer.wordpress.org/reference/hooks/admin_post_action/). It might also be possible to use $_Session

cnohall:
I think I figured out a way that doesn't require _options_. We can use JavaScript _sessionStorage_. It seems to do the job. However, the transition seems a bit laggy. Let me know what you think!

shivaenigma:
> I think I figured out a way that doesn't require _options_. We can use JavaScript _sessionStorage_. It seems to do the job. However, the transition seems a bit laggy. Let me know what you think!

Too complex. Is it possible using POST parameters or hidden form variables ?

blockonomics-woocommerce.php
```diff
@@ -207,30 +181,83 @@ function get_callback_url()
...
                 document.generateSecretForm.submit();
             }
+            function check_form(tab) {
+                const urlParams = new URLSearchParams(window.location.href);
+                const currentTab = urlParams.get('tab') ?? 'settings';
+                if (currentTab == tab){
+                    return;
+                }
+                if (document.getElementById('blockonomics_form_updated').value == 'true' || document.getElementById('blockonomics_api_updated').value == 'true'){
+                    if (currentTab === 'settings'){
```
DarrenWestwood, line 220: 1) One solution is to save the form, using vanilla javascript or ajax request, before redirecting the user
```
if (currentTab === 'settings'){
    if(validateBlockonomicsForm()){
        var xhr = new XMLHttpRequest();
        xhr.open("POST", "options.php"); 
        xhr.onload = function(event){ 
            window.location.href = "options-general.php?page=blockonomics_options&tab=currencies";
        }; 
        var formData = new FormData(document.myform); 
        xhr.send(formData);
    }
} else {
    var xhr = new XMLHttpRequest();
    xhr.open("POST", "options.php"); 
    xhr.onload = function(event){ 
        window.location.href = "options-general.php?page=blockonomics_options&tab=settings";
    }; 
    var formData = new FormData(document.myform); 
    xhr.send(formData);
}
```
2) We could also use a hidden form and submit it, the same as how generateSecret works. We would be required to POST all the form values and save them ourselves.
```
if (currentTab === 'settings'){
    if(validateBlockonomicsForm()){
        document.saveFormAndRedirect.submit();
    }
} else {
    document.saveFormAndRedirect.submit();
}
```
3) We can remove autosaving unsaved changes during _tab changes_, as this is the cause of the added complexity.

shivaenigma:
>One solution is to save the form, using vanilla javascript or ajax request, before redirecting the user

I like this. Lets implement this

shivaenigma:
```
if (currentTab === 'settings'){
    if(validateBlockonomicsForm()){
        var xhr = new XMLHttpRequest();
        xhr.open("POST", "options.php"); 
        xhr.onload = function(event){ 
            window.location.href = "options-general.php?page=blockonomics_options&tab=currencies";
        }; 
        var formData = new FormData(document.myform); 
        xhr.send(formData);
    }
} else {
    var xhr = new XMLHttpRequest();
    xhr.open("POST", "options.php"); 
    xhr.onload = function(event){ 
        window.location.href = "options-general.php?page=blockonomics_options&tab=settings";
    }; 
    var formData = new FormData(document.myform); 
    xhr.send(formData);
}

Note the code redundancy. This could easily be one function with parameter. We should remove similar code redundancy in other code that we wrote that deals with both the tabs
```

cnohall:
Sounds good, I'll implement it and at the same time I'll look at ways of removing code redundancy that deals with both tabs.

cnohall:
The vanilla JavaScript solution of saving the form and then redirecting the user has been implemented. I also simplified some functions and removed code redundancy.

shivaenigma:
Hi @DarrenWestwood  and @cnohall  lets wrap this up in next two days. It has taken a lot time for various reasons

DarrenWestwood (approved):
The code looks good and is functioning well.

blockonomics-woocommerce.php
```diff
@@ -207,28 +207,59 @@ function get_callback_url()
...
+                }
+            }
+            function save_form_then_redirect(tab) {
+                const xhr = new XMLHttpRequest();
+                xhr.open("POST", "options.php"); 
+                xhr.onload = function(event){ 
+                    window.location.href = "options-general.php?page=blockonomics_options&tab="+tab;
+                }; 
+                const formData = new FormData(document.myform); 
+                xhr.send(formData);
```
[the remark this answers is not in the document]
cnohall: Done.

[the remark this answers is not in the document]
DarrenWestwood: The change clears the form before it is saved. Move this to after `xhr.send(formData);` in order to fix the settings saving.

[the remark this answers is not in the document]
cnohall: Thanks, now it's fixed.

DarrenWestwood (approved):

PR #238, opened 2021-04-09
Fixing database error

cnohall:
This PR is created to fix [this error](https://wordpress.org/support/topic/showing-sql-error-from-query-monitor/#post-14285132). I was able to recreate the error and has done some small changes in order to fix it.

There are mainly two things that have changed. The register_activation_hook() has been introduced. It is true that this is not called during an update. However, it seems to be necessary during plugin initiation. 

$sql _CREATE TABLE IF NOT EXISTS_ has been changed to _CREATE TABLE_. This is the way it is structured in the [WordPress documentation](https://codex.wordpress.org/Creating_Tables_with_Plugins). It seems to be the case that dbDelta() does not require _IF NOT EXISTS_, but will instead discern this itself.

As part of testing I've deleted, installed, upgraded the database, with and without orders. It seems to work as it should,

Please review this PR.

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -495,7 +495,7 @@ function blockonomics_create_table() {
 
     $table_name = $wpdb->prefix . 'blockonomics_orders';
     $charset_collate = $wpdb->get_charset_collate();
-    $sql = "CREATE TABLE IF NOT EXISTS $table_name (
+    $sql = "CREATE TABLE $table_name (
         order_id int NOT NULL,
         status int NOT NULL,
         crypto varchar(3) NOT NULL,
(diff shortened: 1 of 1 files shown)


shivaenigma:
> There are mainly two things that have changed. The register_activation_hook() has been introduced. It is true that this is not called during an update. However, it seems to be necessary during plugin initiation.

What is the reason of doing this change and changing the update logic ? As we know plugins_loaded is called on each page so this is really uncessary

It is not enough to just address the real reason of [this error](https://wordpress.org/support/topic/showing-sql-error-from-query-monitor/#post-14285132)? Changing CREATE IF NOT EXISTS TO CREATE  should be sufficient.

cnohall:
If only CREATE IF NOT EXISTS TO CREATE is changed it results in a new database related error. It seems to have to 
image.

Using register_activation_hook() to fire blockonomics_create_table() resolved the error given in the image above.

cnohall:
If only CREATE IF NOT EXISTS TO CREATE is changed it results in a new database related error. It seems to have to 
image.

Using register_activation_hook() to fire blockonomics_create_table() resolved the error given in the image above.

shivaenigma:
Looking good. Will merge after some testing

Issue #233, opened 2021-02-16
Simplify callback checking in TestSetup


blockonomics:
Callback checking in *set_callback_if_needed()* etc methods is still very complex. We need to simplify the logic.

### Test Cases
**Server callback** _https://www.testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_
**Wp callback** _https://www.testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_ 
**Result** _Exact match No changes required test setup_ 

**Server callback** _https://www.testbtcshop.com/wc-api/WC_Gateway_Blockonomics/?secret=91ed0011fecbad319a365f036dd7f81638125e42_
**Wp callback** _https://www.testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_ 
**Result** _Replace callback_ 

**Server callback** _https://testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_
**Wp callback** _https://www.testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_ 
**Result** _Replace callback_ 

**Server callback** _http://testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=8953a0f219d12c722e286798b126_
**Wp callback** _https://www.testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_ 
**Result** _Replace callback_ 

**Server callback** _http://testmyshopdirect.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_
**Wp callback** _https://www.testbtcshop.com?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb_ 
**Result** _Multiple Callbacks Error_ 

### Summary of algorithm
- First check if WP callback is exact match of an existing callback. Then do nothing
- Else replace callback with *exact domain name match*
- Else give error

blockonomics:
This has to work properly when merchant has callbacks from multiple shops say
[...]

PR #239, opened 2021-04-13
Simplify callback logic

cnohall:
This is connected to issue #233 (Simplify callback checking in TestSetup). 

It follows the logic:

- First check if WP callback is exact match of an existing callback. Then do nothing
- Else replace callback with exact domain name match
- Else give error

This still need to be improved to work when merchant has set up multiple shops.

### DIFF
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -157,15 +157,15 @@ class Blockonomics
             $server_callback_url = isset($one_response->callback) ? $one_response->callback : '';
             $server_base_url = preg_replace('/https?:\/\//', '', $server_callback_url);
             $xpub = isset($one_response->address) ? $one_response->address : '';
-            // No callback
             if(!$server_callback_url){
+                // No callback
                 $available_xpub = $xpub;
-            // Exact match
             }else if($server_callback_url == $wordpress_callback_url){
+                // Exact match
                 return '';
             }
-            // Partial Match - Only secret or protocol differ
             else if(strpos($server_callback_url, $api_url) === 0 || strpos($server_base_url, $base_url) === 0 ){
+                // Partial Match - Only secret or protocol differ
                 $this->update_callback($wordpress_callback_url, $crypto, $xpub);
                 return '';
             }
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -131,9 +131,7 @@ class Blockonomics
     public function check_get_callbacks_response_body ($response, $crypto){
         $error_str = '';
         $response_body = json_decode(wp_remote_retrieve_body($response));
-        $callback_secret = get_option('blockonomics_callback_secret');
-        $api_url = WC()->api_request_url('WC_Gateway_Blockonomics');
-        $wordpress_callback_url = add_query_arg('secret', $callback_secret, $api_url);
+
         //if merchant doesn't have any xPubs on his Blockonomics account
         if (!isset($response_body) || count($response_body) == 0)
         {
@@ -142,14 +140,17 @@ class Blockonomics
         //if merchant has at least one xPub on his Blockonomics account
         elseif (count($response_body) >= 1)
         {
-            $error_str = $this->examine_server_callback_urls($response_body, $wordpress_callback_url, $api_url, $crypto);
+            $error_str = $this->examine_server_callback_urls($response_body, $crypto);
         }
         return $error_str;
     }
 
     // checks each existing xpub callback URL to update and/or use
-    public function examine_server_callback_urls($response_body, $wordpress_callback_url, $api_url, $crypto)
+    public function examine_server_callback_urls($response_body, $crypto)
     {
+        $callback_secret = get_option('blockonomics_callback_secret');
+        $api_url = WC()->api_request_url('WC_Gateway_Blockonomics');
+        $wordpress_callback_url = add_query_arg('secret', $callback_secret, $api_url);
         $base_url = preg_replace('/https?:\/\//', '', $api_url);
         $available_xpub = '';
         $partial_match = '';
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -147,57 +147,37 @@ class Blockonomics
         return $error_str;
     }
 
+    // checks each existing xpub callback URL to update and/or use
     public function examine_server_callback_urls($response_body, $wordpress_callback_url, $api_url, $crypto)
     {
-        //Scenario 1: Exact match                       =>      return $error_str;
-        //Scenario 2: All but secret matches            =>      update_callback(), then return $error_str;
-        //Scenario 3: Scenario 1 and 2 failed           =>      Use an empty callback, if present.
-        //Scenario 4: Scenario 1, 2, and 3 failed       =>      Use matching callback with different protocol, if present.
-        //Scenario 5: All above failed                  =>      Multiple callback error: Please add a new store with valid xpub
-
-        //prepare variables
         $base_url = preg_replace('/https?:\/\//', '', $api_url);
-        $matching_callback = false;
-        $different_protocols = false;
-        $different_protocol_xpub = '';
         $available_xpub = '';
         $error_str = '';
 
-        //go through all xpubs on the server and examine their callback url
+        //Go through all xpubs on the server and examine their callback url
         foreach($response_body as $one_response){
             $server_callback_url = isset($one_response->callback) ? $one_response->callback : '';
             $xpub = isset($one_response->address) ? $one_response->address : '';
-            //if the xpub doesn't have a callback => store available xpub
+            // No callback
             if(!$server_callback_url){
                 $available_xpub = $xpub;
-            //Exact match => return
+            // Exact match
             }else if($server_callback_url == $wordpress_callback_url){
-                return $error_str;
-            //All but secret match => update_callback and return
-            }else if(strpos($server_callback_url, $api_url) === 0){
-                $this->update_callback($wordpress_callback_url, $crypto, $xpub);
-                return $error_str;
+                return '';
             }
-            //matching url but the protocols are different => store values and only update as last resort
+            // Partial Match - Only secret or protocol differ
             else if(strpos($server_callback_url, $base_url)){
-                $different_protocols = true;
-                $different_protocol_xpub = $xpub;
+                $this->update_callback($wordpress_callback_url, $crypto, $xpub);
+                return '';
             }
         }
-
-        //Scenario 1 (Exact match) and 2 (All but secret matches) failed => use available xpub if we have one  
+        // Use the available xpub
         if($available_xpub){
             $this->update_callback($wordpress_callback_url, $crypto, $available_xpub);
-            $matching_callback = true;
-        //Scenario 1 (Exact match) and 2 (All but secret matches) failed, also no available_xpub => use callback with different protocol
-        } elseif ($different_protocols){
-            $this->update_callback($wordpress_callback_url, $crypto, $different_protocol_xpub);
-            $matching_callback = true;
+            return '';
         }
-        //No match and no empty callback
-        if(!$matching_callback){
-            $error_str = __("Multiple callback error: Please add a new store with valid xpub", 'blockonomics-bitcoin-payments');
-        }
+        // No match and no empty callback
+        $error_str = __("Multiple callback error: Please add a new store with valid xpub", 'blockonomics-bitcoin-payments');
         return $error_str;
     }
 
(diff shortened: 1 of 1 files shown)


php/Blockonomics.php
```diff
@@ -131,56 +131,34 @@ public function check_get_callbacks_response_code($response){
...
             if(isset($response_body[0])){
                 $response_callback = isset($response_body[0]->callback) ? $response_body[0]->callback : '';
-                $response_address = isset($response_body[0]->address) ? $response_body[0]->address : '';
+                $xpub = isset($response_body[0]->address) ? $response_body[0]->address : '';
             }
-            $response_callback_without_schema = preg_replace('/https?:\/\//', '', $response_callback);
-            if(!$response_callback || $response_callback == null)
-            {
-                //No callback URL set, set one 
-                $this->update_callback($callback_url, $crypto, $response_address);
```
DarrenWestwood, line 156: This code automatically sets the callback url if the xpub has an empty address callback url and is missing from the new logic.

DarrenWestwood:
We can adapt this code to work for multiple websites by using two flags. eg. `$matching_callback` and `$available_xpub`
```
$error_str = '';
if (!isset($response_body) || count($response_body) == 0){
  $error_str = __('You have not entered an xPub', 'blockonomics-bitcoin-payments');
}else if (count($response_body) >= 1) {
  $matching_callback = false;
  $available_xpub = '';
  $base_url = preg_replace('/https?:\/\//', '', $base_url);
  foreach($response_body as $xpub){
      //Empty callback
      if(!$xpub->callback){
        $available_xpub = $xpub->callback;
      //Exact match
      }else if($xpub->callback == $callback_url){
        $matching_callback = true;
        break;
      //Domains have exact match but other details are different
      }else if(strpos($xpub->callback, $base_url)){
        $this->update_callback($callback_url, $crypto, $xpub);
        $matching_callback = true;
        break;
      }
  }
  if(!$matching_callback && $available_xpub){
    $this->update_callback($callback_url, $crypto, $available_xpub);
    $matching_callback = true;
  }
  if(!$matching_callback){
      $error_str = __("You have an existing callback URL", 'blockonomics-bitcoin-payments');
  }
}
return $error_str;
```

blockonomics:
Only matching domain will not work as we have to think about [multisite also](https://github.com/blockonomics/woocommerce-plugin/issues/233#issuecomment-833610158). It is better to have an excel sheet of cases with expected output for regression

php/Blockonomics.php
cnohall: True, thanks. It's now been fixed.

cnohall:
[Google docs spreadsheet](https://docs.google.com/spreadsheets/d/1Y5FsCPa6RsMOOBkHUCcgdcwQNUdN5UCJvJ3vz0asVh4/edit?usp=sharing) that holds the logic and expected behaviour of the callback

php/Blockonomics.php
```diff
@@ -131,59 +131,78 @@ public function check_get_callbacks_response_code($response){
...
-             if(preg_replace('/https?:\/\//', '', $res_obj->callback) == $callback_url_without_schema)
-                $error_str = "";
-        }  
+
+        //Scenario 1 (Exact match) and 2 (All but secret matches) failed => use available xpub if we have one  
+        if($available_xpub){
+            $this->update_callback($wordpress_callback_url, $crypto, $available_xpub);
+            $matching_callback = true;
+        //Scenario 1 (Exact match) and 2 (All but secret matches) failed, also no available_xpub => use callback with different protocol
+        } elseif ($different_protocols){
```
DarrenWestwood, line 193: If there is an _available_xpub_ and an xpub with a callback URL that has _different_protocols_, the callback with a different protocol should be updated and used. In this case, the available_xpub is used, which can lead to `xpub1 - http://store.com/?secret=123` and `xpub2 - https://store.com/?secret=123` instead of updating xpub1.

In any scenario, if a partly matching domain is found it should be updated and test setup return success. To keep the code simple we only need 3 checks
```
  // No callback
  if(!$server_callback_url){
    $available_xpub = $xpub;
  }
  // Exact match
  }else if($server_callback_url == $wordpress_callback_url){
    return '';
  }
  // Partial Match - Only secret or protocol differ etc
  else if(strpos($server_callback_url, $api_url) === 0 || strpos($server_callback_url, $base_url)){
    $this->update_callback($wordpress_callback_url, $crypto, $xpub);
    return '';
  }
}

// Use the available xpub
if($available_xpub){
  $this->update_callback($wordpress_callback_url, $crypto, $available_xpub);
  return '';
}
// No match and no empty callback
$error_str = __("Multiple callback error: Please add a new store with valid xpub", 'blockonomics-bitcoin-payments');
return $error_str;
```
This removes `$different_protocols` and `$different_protocol_xpub` as we update and return immediately. We can also remove `$matching_callback`  and `$error_str = '';` as we return an empty string instead and only use `$error_str` for the error.

php/Blockonomics.php
```diff
@@ -131,59 +131,78 @@ public function check_get_callbacks_response_code($response){
...
-                //No callback URL set, set one 
-                $this->update_callback($callback_url, $crypto, $response_address);
+            $error_str = $this->examine_server_callback_urls($response_body, $wordpress_callback_url, $api_url, $crypto);
+        }
+        return $error_str;
+    }
+
+    public function examine_server_callback_urls($response_body, $wordpress_callback_url, $api_url, $crypto)
+    {
+        //Scenario 1: Exact match => return $error_str;
```
DarrenWestwood, line 152: These comments should be unnecessary if the relevant code is commented on correctly and the code understandable. We can add a single comment to `examine_server_callback_urls()` to explain the function checks each existing xpub callback URL to update and/or use.

php/Blockonomics.php
cnohall: I agree, if we don't check the protocols the algorithm becomes more streamlined.

php/Blockonomics.php
cnohall: I've reduced the comments into one single line.

cnohall:
The callback has been improved to handle multiple callbacks and simplified with clearer logic.

php/Blockonomics.php
DarrenWestwood: We can still check if only the protocol differs to constitute a Partial Match. Different secret or different protocol can be handled the same. ie update and return success:
```
// Partial Match - Only secret or protocol differ etc
else if(strpos($server_callback_url, $api_url) === 0 || strpos($server_callback_url, $base_url)){
  $this->update_callback($wordpress_callback_url, $crypto, $xpub);
  return '';
}
```

php/Blockonomics.php
```diff
@@ -131,59 +131,58 @@ public function check_get_callbacks_response_code($response){
...
+        }
+        return $error_str;
+    }
+
+    // checks each existing xpub callback URL to update and/or use
+    public function examine_server_callback_urls($response_body, $wordpress_callback_url, $api_url, $crypto)
+    {
+        $base_url = preg_replace('/https?:\/\//', '', $api_url);
+        $available_xpub = '';
+        $error_str = '';
```
DarrenWestwood, line 155: `$error_str = '';` can also be removed as we directly return `return '';` in case of success.

php/Blockonomics.php
cnohall: Sure, just thought that it was easier to keep it to `strpos($server_callback_url, $base_url)`, since it will catch both different secrets and different protocols. The end result should be the same?

However, I've re-introduced the `strpos($server_callback_url, $api_url) === 0` check again.

php/Blockonomics.php
cnohall: True, now it's been removed!

php/Blockonomics.php
```diff
@@ -131,59 +131,56 @@ public function check_get_callbacks_response_code($response){
...
-                    //Just force Update_callback on server
-                    $this->update_callback($callback_url, $crypto, $response_address);
-                }
-                else
-                {
-                    $error_str = __("You have an existing callback URL", 'blockonomics-bitcoin-payments');
-                }
-                
+            // Partial Match - Only secret or protocol differ
+            else if(strpos($server_callback_url, $api_url) === 0 || strpos($server_callback_url, $base_url)){
```
DarrenWestwood, line 167: The code is clean and readable. Checking for a Partial Match requires further work as `strpos($server_callback_url, $base_url)` updates the callback URL incorrectly if a different subdomain is used.

Test case (different subdomain):
- Blockonomics Server: http://subdomain.shop.com/wc-api/WC_Gateway_Blockonomics/?secret=123
- WordPress URL: http://shop.com/wc-api/WC_Gateway_Blockonomics/?secret=123

The URL should not update in this case as subdomain.shop.com can be a different store to shop.com

cnohall: You're right, it was fairly easy to fix. I added the a variable `$wp_base_url` and make sure that $base_url start at the same position. That way no subdomain can be "hiding" on the server without getting detected and being called out as a multiple callback error.

php/Blockonomics.php
```diff
@@ -131,59 +131,57 @@ public function check_get_callbacks_response_code($response){
...
         }
-        else 
-        {
-            $error_str = __("You have an existing callback URL", 'blockonomics-bitcoin-payments');
-            // Check if callback url is set
-            foreach ($response_body as $res_obj)
-             if(preg_replace('/https?:\/\//', '', $res_obj->callback) == $callback_url_without_schema)
-                $error_str = "";
-        }  
+        // Use the available xpub
```
DarrenWestwood, line 173: Test Case 20 will possibly update to the same callback url, which can create problems if the same url matches two xpubs. We should reinclude the $partial_match flag and only update the partial_match after checking all the callback URLs do not contain an exact match.
```
if($partial_match || $available_xpub){
  $update_xpub = $partial_match ? $partial_match : $available_xpub;
  $this->update_callback($wordpress_callback_url, $crypto, $update_xpub);
  return '';
}
```

DarrenWestwood (commented):
Checking each callback URL for an exact or partial match is now working well with the test cases.

php/Blockonomics.php
```diff
@@ -131,59 +131,57 @@ public function check_get_callbacks_response_code($response){
...
+    public function examine_server_callback_urls($response_body, $wordpress_callback_url, $api_url, $crypto)
+    {
+        $base_url = preg_replace('/https?:\/\//', '', $api_url);
+        $available_xpub = '';
+        //Go through all xpubs on the server and examine their callback url
+        foreach($response_body as $one_response){
+            $server_callback_url = isset($one_response->callback) ? $one_response->callback : '';
+            $server_base_url = preg_replace('/https?:\/\//', '', $server_callback_url);
+            $xpub = isset($one_response->address) ? $one_response->address : '';
+            // No callback
```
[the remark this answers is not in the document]
cnohall: I prefer PSR-2, so I took the freedom to change it to PSR-2.

php/Blockonomics.php
cnohall: Smart solution, I updated and pushed it, I'll do some testing tomorrow morning

cnohall: Does the algorithm need to account for if there are two partial matches?

Example:
```
server callback 1:  http://shop.home/?wc-api=WC_Gateway_Blockonomics&secret=asdasdkljlkjasd0912312a
server callback 2:  https://shop.home/?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb
wordpress callback: http://shop.home/?wc-api=WC_Gateway_Blockonomics&secret=f5e53a0f219d12c722e286798b126ffed29ee1eb
```
The current algorithm will pick one of the two server callbacks, whichever one is closer to the end of the array, and update it.

If we don't want to treat protocol and secret partial matches differently, I suggest that we only use `strpos($server_base_url, $base_url) === 0` for checking partial matches. It will catch both urls with different _protocols_ and _secrets_. Thanks to checking the position of the match it makes sure that websites with extra subdomains aren't falsely flagged. 

If we want to check for two partial matches, perhaps the smartest option is to notify the user that he has two URLs that closely resemble each other and ask him to take action; since it doesn't seem to happen often we might not want to complicate the logic for it.

DarrenWestwood (approved):
The PR changes are neat and working correctly. Some of the Test cases are required to be updated in the Testing Doc:
1. Test Case 12 Server Callback 2 should not contain www or the empty callback url should be updated
2. Test Case 15 Callback 2 should not contain www or the multiple callback error should be displayed
3. Test Cases 21 and 22 should both return multiple callback error

cnohall:
Sounds good, I've updated the Testing Doc.

Issue #247, opened 2021-05-31
BCH New adddress gen error


shivaenigma:
BCH has a different [error message](https://github.com/blockonomics/woocommerce-plugin/blob/7ec3f6654e8bef08facf50df6acbf232b151d0d5/templates/blockonomics_checkout.php#L38). This was put before we had BCH test setup working and now can be removed

Both BCH and BTC can show the same error message _to click on Settings > Blockonomics > Test Setup_ and code should be refactored / reused properly

cnohall:
I will look into it and make the needed adjustments.

cnohall:
Question about the wording. In the current format we have this message for gap limit errors:
![image](https://user-images.githubusercontent.com/36883813/120170906-6ac42d80-c23c-11eb-982c-826ceae5f780.png)

I suggest that we remove _Note to webmaster:_ or in other ways update the message, since two colons in the same sentence looks a bit strange. This is how it looks like if we remove _Note to webmaster:_

![image](https://user-images.githubusercontent.com/36883813/120170802-4ec08c00-c23c-11eb-8640-e3c2e1d271d8.png)

shivaenigma:
There is a rational behind *Note to webmaster* . We can of course change the punctuation/text etc. Here is the basic idea:

- Customer doing checkout should not be shown payment processor errors. This confuses him and he thinks he did something wrong during the checkout. Again since, this is related to bitcoin it further confuses him
- Thus the error shown is directed only to *webmaster* and he is instructed to use *Test Setup* to further understand the error
- In case of Gap limit error, the problem is that **Test Setup** will never show this, because it always call the API with *test flag* on .We haven't thought of the correct solution to this, so meanwhile we are just showing during checkout. This error is of course quite rare

cnohall:
Thanks for the explanation, makes sense!

PR #249, opened 2021-06-01
BCH New adddress gen error

cnohall:
This PR is related to issue #247. 

BTC and BCH both show the same error and duplicate code has been removed and refactored. 

One small https://github.com/blockonomics/woocommerce-plugin/issues/247#issuecomment-851350463 is that I would like to remove the _Note to webmaster:_ part of the gaplimit error message. Let me know what you think.


DarrenWestwood (approved):
The changes are clean and working correctly. Let's replace `Note to webmaster:` with `Note to webmaster -` as the message is still relevant to the site admin and should remain.

Issue #241, opened 2021-05-01
🚨 Potential Cross-site Scripting (XSS) - Reflected (CWE-79)


huntr-helper:
👋 Hello, @blockonomics - a potential high severity Cross-site Scripting (XSS) - Reflected (CWE-79) vulnerability in your repository has been disclosed to us.

#### Next Steps

1️⃣  Visit **https://huntr.dev/bounties/1-other-blockonomics/woocommerce-plugin** for more advisory information.

2️⃣  **[Sign-up](https://huntr.dev/)** to validate or speak to the researcher for more assistance.

3️⃣ Propose a patch or outsource it to our community - whoever fixes it gets paid.

✏️ **NOTE**: If we don't hear from you in *14 days*, we will proactively source a fix for this vulnerability (and open a PR) to ensure community safety.

---

#### Confused or need more help?

- Join us on our **[Discord](https://huntr.dev/discord)** and a member of our team will be happy to help! 🤗

- Speak to a member of our team: @JamieSlome

---

*This issue was automatically generated by [huntr.dev](https://huntr.dev) - a bug bounty board for securing open source code.*

DarrenWestwood:
[esc_attr()](https://developer.wordpress.org/reference/functions/esc_attr/) should be used instead of sanitize_text_field() to secure the attribute correctly.

shivaenigma:
Should we fix this in other places as well where we are using sanitize_text_field ?

DarrenWestwood:
All other places are correctly sanitizing or escaping input data. It is recommended to always sanitize, validate, or escape request data to avoid sending incorrect data through the system.
Added sanitization to requests in the select crypto page, nojs checkout page, and admin options.
https://github.com/blockonomics/woocommerce-plugin/pull/250

As an additonal precaution, all data output in the templates/admin can be escaped when echoing.

PR #250, opened 2021-06-01
Fix escaping filter_by request

DarrenWestwood:
Secure filter_by attribute using esc_attr(), Issue #241
Tested the attribute is correctly escaped:
XSS fix


blockonomics:
I am merging this for now. Will review security later

Issue #252, opened 2021-06-23
Crypto Selection Screen Update


shivaenigma:
If feel we can still improve the crypto selection a lot:

1. [Non clickable regions](https://drive.google.com/file/d/1GfMAoeG-E7gTL4QIBv7zBjhBQ7U1-rot/view?usp=sharing)
2. Improve look / feel further
3. Mobile usability must be good. Right now mobile is same as desktop which is wrong. Mobile should change to a [vertical layout](https://prnt.sc/16gwbt8)

cnohall:
Sound good, looking forward to working on this!

PR #253, opened 2021-06-30
Fix crypto selection button edges

cnohall:
This is related to Issue #252.

This PR makes so that the edges of the buttons are clickable. I will create another separate PR with the CSS for responsive buttons (vertical instead of horizontal on mobiles).


blockonomics:
Lets do a bigger PR that addresses all things in #252. I would believe changing the UI look/feel would fix these anyway. Mobile usability is quite important

Issue #251, opened 2021-06-23
Refershing payment screen should reset timer and price


shivaenigma:
If user sees the timer already at say 2 min, as he doesn't have time. He will do this:

1. Press back and go back to crypto selection screen or *Pay with  Bitcoin* Wordpress checkout screen
2. He hopes that a new checkout is created, when he comes back which doesn't happen and he is still at around 2 min
3. WHMCS currently works like this. Address can be resused as it same order number, however timer / price should be refreshed

cnohall:
Sound good, looking forward to working on this!

shivaenigma:
This is a complex issue and here are few reasons why we till now retain page values on refresh:

- Opening the same checkout in a new tab/multiple refreshes may lead to incorrect expected amounts
- Customer can send the link to another person to pay, in which case it is desirable they see the same amount
- Mobile will discard tab and reload it sometimes https://www.chromium.org/chromium-os/chromiumos-design-docs/tab-discarding-and-reloading

However, looking afresh from customer point of view, it is important this behaviour is changed:
- Problems when time [remaining is low](https://github.com/blockonomics/woocommerce-plugin/issues/251#issue-928445347)
- More importantly refresh should always **show the latest bitcoin price** as far as possible. This is fair to both the customer and the merchant. If customer enters the BTC amount in his wallet and the fiat amount is deviating a lot, it causes a problem

Following changes must be implemented:
- Refreshing crypto payment screen will reset time and update price if order is still hasn't received any tx. Address will of course be reused 
- Technical notion of *time_remaining* etc and other uncessary time variables on server/client side can be removed

PR #257, opened 2021-07-05
Reset checkout timer at refresh

cnohall:
Related to Issue #251.

The code has seen small changes, but it should be enough to implement the proposed changes. Every time a user refreshes the page, an API call will be sent to check and update the price, and the timer will be reset.

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -500,7 +500,6 @@ function blockonomics_create_table() {
         status int NOT NULL,
         crypto varchar(3) NOT NULL,
         address varchar(191) NOT NULL,
-        timestamp int,
         satoshi int,
         currency varchar(3),
         value longtext,
@@ -532,7 +531,7 @@ function blockonomics_update_db_check() {
     if ( $installed_ver != $blockonomics_db_version ) {
         $table_name = $wpdb->prefix . 'blockonomics_orders';
         if ($installed_ver < 1.1) {
-            $wpdb->query("ALTER TABLE $table_name DROP time_remaining;");
+            $wpdb->query("ALTER TABLE $table_name DROP COLUMN time_remaining, timestamp;");
         }
         blockonomics_create_table();
     }
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -501,7 +501,6 @@ function blockonomics_create_table() {
         crypto varchar(3) NOT NULL,
         address varchar(191) NOT NULL,
         timestamp int,
-        time_remaining int,
         satoshi int,
         currency varchar(3),
         value longtext,
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -531,7 +531,7 @@ function blockonomics_update_db_check() {
     $installed_ver = get_site_option( 'blockonomics_db_version' );
     if ( $installed_ver != $blockonomics_db_version ) {
         $table_name = $wpdb->prefix . 'blockonomics_orders';
-        if ($blockonomics_db_version > 1.0) {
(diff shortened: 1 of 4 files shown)


DarrenWestwood (commented):
Working as desired.

The time_remaining column should be removed for upgrading users. Architecture for database upgrades can be found here: https://github.com/blockonomics/woocommerce-plugin/blob/master/blockonomics-woocommerce.php#L511-L514

php/Blockonomics.php
```diff
@@ -466,9 +466,8 @@ public function calculate_order_params($order){
                 $price = 1;
             }
             $order['satoshi'] = intval(round(1.0e8*$wc_order->get_total()/$price));
-            $order['timestamp'] = time();
```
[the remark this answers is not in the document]
cnohall: It has now been reinstated

blockonomics-woocommerce.php
```diff
@@ -531,10 +530,10 @@ function blockonomics_update_db_check() {
 
     $installed_ver = get_site_option( 'blockonomics_db_version' );
     if ( $installed_ver != $blockonomics_db_version ) {
-        // Example function to demonstrate table changes between upgrade versions
-        // if ($installed_ver < 1.0) {
-        //     $wpdb->query("ALTER TABLE $table_name DROP transaction;");
-        // }
+        $table_name = $wpdb->prefix . 'blockonomics_orders';
+        if ($blockonomics_db_version > 1.0) {
```
DarrenWestwood, line 534: This code will run again incorrectly when `if ($blockonomics_db_version > 1.1)` is added and the user upgrades from db_version 1.1->1.2.
We should use `if ($installed_ver < 1.1)` as we want to run the upgrades between the installed version and the new db_version. This also makes the code easier to read as the if statement references the matching db_version

cnohall: Ah, that makes more sense. Thanks for the clarification! I've pushed the change

DarrenWestwood (approved):

blockonomics-woocommerce.php
```diff
@@ -531,10 +530,10 @@ function blockonomics_update_db_check() {
 
     $installed_ver = get_site_option( 'blockonomics_db_version' );
     if ( $installed_ver != $blockonomics_db_version ) {
-        // Example function to demonstrate table changes between upgrade versions
-        // if ($installed_ver < 1.0) {
-        //     $wpdb->query("ALTER TABLE $table_name DROP transaction;");
-        // }
+        $table_name = $wpdb->prefix . 'blockonomics_orders';
+        if ($installed_ver < 1.1) {
+            $wpdb->query("ALTER TABLE $table_name DROP time_remaining;");
```
blockonomics, line 535: As far I remember dbDelta will automatically drop the column when `blockonomics_create_table` is called so this is uncessary. Please verify with real table having values

blockonomics-woocommerce.php
```diff
@@ -501,7 +501,6 @@ function blockonomics_create_table() {
         crypto varchar(3) NOT NULL,
         address varchar(191) NOT NULL,
         timestamp int,
```
blockonomics, line 503: What is the use of `timestamp` column now?  If its not being used anymore better to remove this also

blockonomics (changes requested):

blockonomics-woocommerce.php
cnohall: It seems like dbDelta cannot delete columns. ([source](https://wordpress.stackexchange.com/a/76930))

blockonomics-woocommerce.php
cnohall: true, now the `timestamp` has been dropped

cnohall:
The requested changes have been done

PR #259, opened 2021-07-16
Set header as Array instead of string

cnohall:
This PR is small and only contains two changes.

- Version changed from 3.3 => 3.3.1
-  `'Authorization: Bearer ' . $api_key;` is now `array('Authorization' => 'Bearer ' . $api_key);`

It works and follows WordPress' standards.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,8 @@
 == Changelog ==
 
+= 3.3.1  =
+* Minor bug fix
+
 = 3.3  =
 * Minor Security fixes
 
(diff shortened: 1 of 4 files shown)


blockonomics:
Please verify if this breaks anything. The reason this was done is because for one merchant APIKey was returning invalid inspite of being correct. Having plugin WPSSO seems to be effecting this

Also plugin API is in inline [with standards](https://stripe.com/docs/api/authentication)

cnohall:
I'll close this as well

PR #263, opened 2021-08-12
Fix duplicate orders

DarrenWestwood:
Fixes issues with concurrent orders using the same bitcoin address.
- Saves the order to the blockonomics_orders table before updating the post metadata using record_address(). This ensures duplicates are checked before saving the address to the post
- Displays the _Temporary address error_ if SQL error on insert_order(), caused by the duplicate entry.

The fix can be tested using the following steps:
1. Create two new orders on WooCommerce. eg. Orders 1476 and 1477
2. Delete the two new orders from the blockonomics_orders table
3. Run the following curl command after replacing the two order_ids and your WordPress URL, in order to call get_order for both orders at the same time `seq 1476 1477 | xargs -I + -P2 curl 'http://localhost/wordpress/index.php/wc-api/WC_Gateway_Blockonomics?crypto=btc&get_order=+'`
One order receives the new address, the other exits with the _Temporary address error_

### DIFF
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -554,11 +554,14 @@ class Blockonomics
     public function process_order($order_id, $crypto){
         $order = $this->get_order_by_id_and_crypto($order_id, $crypto);
         if ($order) {
+            // Update the existing order info
             $order = $this->calculate_order_params($order);
             $this->update_order($order);
         }else {
+            // Create and add the new order to the database
             $order = $this->create_new_order($order_id, $crypto);
             if (!$this->insert_order($order)) {
+                // insert_order fails if duplicate address found. Ensures no duplicate orders in the database
                 exit(json_encode(array("error"=>"Duplicate Address Error. This is Temporary error, Please try again")));
             }
             $this->record_address($order_id, $crypto, $order['address']);
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -562,7 +562,7 @@ class Blockonomics
             $order = $this->create_new_order($order_id, $crypto);
             if (!$this->insert_order($order)) {
                 // insert_order fails if duplicate address found. Ensures no duplicate orders in the database
-                exit(json_encode(array("error"=>"Duplicate Address Error. This is Temporary error, Please try again")));
+                exit(json_encode(array("error"=>__("Duplicate Address Error. This is Temporary error, Please try again", 'blockonomics-bitcoin-payments'))));
             }
             $this->record_address($order_id, $crypto, $order['address']);
         }
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -559,7 +559,7 @@ class Blockonomics
         }else {
             $order = $this->create_new_order($order_id, $crypto);
             if (!$this->insert_order($order)) {
-                exit(json_encode(array("error"=>"Temporary address error, please try again")));
+                exit(json_encode(array("error"=>"Duplicate Address Error. This is Temporary error, Please try again")));
             }
             $this->record_address($order_id, $crypto, $order['address']);
         }
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -562,7 +562,7 @@ class Blockonomics
             $order = $this->create_new_order($order_id, $crypto);
             if (!$this->insert_order($order)) {
                 // insert_order fails if duplicate address found. Ensures no duplicate orders in the database
-                exit(json_encode(array("error"=>__("Duplicate Address Error. This is Temporary error, Please try again", 'blockonomics-bitcoin-payments'))));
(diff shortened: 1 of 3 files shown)


php/Blockonomics.php
```diff
@@ -487,16 +487,13 @@ public function create_new_order($order_id, $crypto){
             exit(json_encode(array("error"=>$responseObj->response_message)));
         }
         $address = $responseObj->address;
```
blockonomics, line 489: Instead of changing rest of the code, I prefer we check if the address is already existing here and exit with error in the same way we are existing in line 487. 

_Duplicate Address Error.  This is Temporary error, Please try again_

templates/blockonomics_checkout.php
```diff
@@ -39,6 +39,11 @@
         <h2><?=__('Could not generate new Bitcoin Cash address', 'blockonomics-bitcoin-payments')?></h2>
         <p><?=__('Note to webmaster: Please follow the instructions <a href="https://help.blockonomics.co/en/support/solutions/articles/33000253348-bch-setup-on-woocommerce" target="_blank">here</a> to configure BCH payments.', 'blockonomics-bitcoin-payments')?></p>
       </div>
+      <!-- Duplicate Address Error -->
+      <div id="address-error-duplicate" ng-show="address_error_duplicate" ng-cloak>
```
blockonomics, line 43: We are uncesssary creating multiple divs for errrors. I prefer we have a div _customer_facing_error_

app.js can match is string contains gap limit or temporary error and set this div

php/Blockonomics.php
DarrenWestwood: If both addresses are returned close to the same time, the 1st matching address may not be saved to the database yet. 

It is best to let the database handle checking for duplicates when saving the new address, which is currently performed once the complete order is available:
https://github.com/blockonomics/woocommerce-plugin/pull/263/files#diff-a3307e680eb76b9d93556c9ae2e5dc134b3526246480730fb8ddf49186da29ebR561

Another option to reduce the code impact is to update the database to use KEY orderkey (order_id, address, crypto), as currently, the 2nd order id is replacing the 1st in the database due to using `replace` to insert or replace:
https://github.com/blockonomics/woocommerce-plugin/blob/master/blockonomics-woocommerce.php#L485

templates/blockonomics_checkout.php
DarrenWestwood: I will update the Duplicate Address Error message. 

It will be better to update to the single customer_facing_error in a separate PR as the translation strings will require passing to js:
https://make.wordpress.org/core/2018/11/09/new-javascript-i18n-support-in-wordpress/

php/Blockonomics.php
blockonomics: >If both addresses are returned close to the same time, the 1st matching address may not be saved to the database yet.
I agree with this

I am unsure how this will give [an error](https://github.com/blockonomics/woocommerce-plugin/pull/263/files#diff-a3307e680eb76b9d93556c9ae2e5dc134b3526246480730fb8ddf49186da29ebR561) when an existing addresses is inserted. The basic problem is we are lacking UNIQUE tag on address column

A select to check uniqueness immediately followed by wp update/replace should work in 99.9% of cases, but it still open to some [race conditions](https://stackoverflow.com/questions/48475909/if-record-exists-when-using-wpdb-insert-dont-insert-and-dont-throw-failure) 

The better solution is to create a new address column having UNIQUE tag and use this column from now.  We cannot put UNIQUE on existing column as it may already have non uniques. Then if _wp->replace_ gives exception we handle it and then show the error

DarrenWestwood: Currently, there may be two orders with the same address in the post meta, however, the database will not have any duplicate addresses as there is already a unique key on the address:
https://github.com/blockonomics/woocommerce-plugin/blob/master/blockonomics-woocommerce.php#L484 ​
This leads to the order_id being replaced by the second order with the same address as we are using replace to insert or replace:
https://github.com/blockonomics/woocommerce-plugin/blob/master/php/Blockonomics.php#L553

[This will give an error](https://github.com/blockonomics/woocommerce-plugin/pull/263/files#diff-a3307e680eb76b9d93556c9ae2e5dc134b3526246480730fb8ddf49186da29ebR561) when an existing address is inserted because the insert already checks for a duplicate key.

In order to get the existing `replace` to work instead of using the insert, a unique key should be placed on order_id, address, and crypto, as we currently only use the address and crypto:
https://github.com/blockonomics/woocommerce-plugin/blob/master/blockonomics-woocommerce.php#L485

blockonomics: Ok understood. Apologies. Wrong understanding by me. 

It is actually well written code
Please add code comments to explain workflow/motivation as later it will be difficult to understnad

templates/blockonomics_checkout.php
blockonomics: Can you finish this so that we can close the PR

DarrenWestwood: Refactored to single customer_facing_error in 463c4eb

PR #270, opened 2022-03-29
(update) Encrypt Order IDs in checkout

### LINKED ISSUE #237: Order id should be encrypted like WHMCS plugin
It may be possible to view other order information by changing order id in order info URLs. We should either encrypt order information or use a separate column in table with a unique id

thisisayush:
Fixes #237 

Encrypts Order ID for various endpoints to prevent leakage of order info by incrementing IDs

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -26,7 +26,8 @@ function CheckoutController($scope, $interval, Order, $timeout, Url) {
     $scope.no_display_error = true;
     $scope.copyshow = false;
     //fetch url params
-    $scope.order_id = Url.get_parameter_by_name("show_order");
+    $scope.order_id_hash = Url.get_parameter_by_name("show_order");
+    $scope.order_id = null
 
     check_blockonomics_order();
 
@@ -60,14 +61,14 @@ function CheckoutController($scope, $interval, Order, $timeout, Url) {
                 ws.close();
                 $interval(function() {
                     //Redirect to order confirmation page if message from socket
-                    window.location = Url.get_wc_endpoint({'finish_order' : $scope.order_id});
+                    window.location = Url.get_wc_endpoint({'finish_order' : $scope.order_id_hash});
                 //Wait for 2 seconds for order status to update on server
                 }, 2000, 1);
             }
         }
         else if ($scope.order.status >= 0){
           //Goto order confirmation as payment is already in process or done
-          window.location = Url.get_wc_endpoint({'finish_order' : $scope.order_id});
+          window.location = Url.get_wc_endpoint({'finish_order' : $scope.order_id_hash});
         }
 
     }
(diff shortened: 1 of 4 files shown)


js/app.js
```diff
@@ -26,7 +26,8 @@ function CheckoutController($scope, $interval, Order, $timeout, Url) {
     $scope.no_display_error = true;
     $scope.copyshow = false;
     //fetch url params
-    $scope.order_id = Url.get_parameter_by_name("show_order");
+    $scope.order_id_hash = Url.get_parameter_by_name("show_order");
+    $scope.order_id = null
```
DarrenWestwood, line 30: This line is not required as the order_id is set after get_order

thisisayush:
Updated.

DarrenWestwood (commented):
Encryption/Decryption is working as desired on get_order and finish_order. 
Awaiting fix to check for invalid hash e.g. 
valid: get_order=dcf66ee053fcd8bee1dcd475af97c4d4
invalid: get_order=dcf66

php/WC_Gateway_Blockonomics.php
```diff
@@ -114,16 +114,38 @@ public function handle_requests()
...
+        } else if ($get_order) {
+            $order_hash = $get_order;
+        }
+        
+        $order_id = "";
+
+        if ($order_hash) {
+            $order_id = $blockonomics->decrypt_hash($order_hash);
+            if (empty(wc_get_order($order_id))) {
+                echo "Invalid Order";
```
thisisayush, line 134: @DarrenWestwood What should be the output in such cases?

php/WC_Gateway_Blockonomics.php
```diff
@@ -114,16 +114,38 @@ public function handle_requests()
...
+            $order_hash = $finish_order;
+        } else if ($get_order) {
+            $order_hash = $get_order;
+        }
+        
+        $order_id = "";
+
+        if ($order_hash) {
+            $order_id = $blockonomics->decrypt_hash($order_hash);
+            if (empty(wc_get_order($order_id))) {
```
DarrenWestwood, line 133: The check for woocommerce order should be moved into decrypt_hash function to ensure this check always happens after decrypting the hash, such as in blockonomics_nojs_checkout.php.

DarrenWestwood, line 133: The check for woocommerce order should be moved into decrypt_hash function to ensure this check always happens after decrypting the hash, such as in blockonomics_nojs_checkout.php.

php/WC_Gateway_Blockonomics.php
```diff
@@ -114,16 +114,34 @@ public function handle_requests()
         include_once 'Blockonomics.php';
         $blockonomics = new Blockonomics;
 
+        $order_hash = "";
```
DarrenWestwood, line 117: The order_hash flag is adding lots of unnecessary code. It is also not required to decrypt the order hash with select_crypto. Please refactor decrypting the hash to where it is required:
```
else if ($show_order && $crypto) {
    $order_id = $blockonomics->decrypt_hash($order_hash);
    $blockonomics->load_checkout_template($order_id, $crypto);
}else if ($select_crypto) {
    $blockonomics->load_blockonomics_template('crypto_options');
}else if ($finish_order) {
    $order_id = $blockonomics->decrypt_hash($order_hash);
    $blockonomics->redirect_finish_order($order_id);
}else if ($get_order && $crypto) {
    $order_id = $blockonomics->decrypt_hash($order_hash);
    $blockonomics->get_order_info($order_id, $crypto);
}
```

thisisayush: Done

DarrenWestwood (approved):

PR #271, opened 2022-03-29
(add) Fix for CURCY Multi Currency Selector

thisisayush:
Hides the Currency Switcher in Blockonomics Order Pages when [Woocommerce Multi-Currency Plugin (CURCY)](https://villatheme.com/extensions/woo-multi-currency/) is enabled to prevent a change of currency in the middle of the checkout process.


blockonomics:
Very low priority as usecase is rare. Don't merge for now.

PR #273, opened 2022-05-23
Minor fixes

blockonomics:


templates/blockonomics_nojs_checkout.php
```diff
@@ -55,7 +54,7 @@
                 </div>
 
                 <div>
-                  <a href="<?php echo $check_payment_url;?>">Click here if already paid</a>
+                  <a href="">Click here if already paid</a>
```
DarrenWestwood, line 57: Changes are working well. It should be noted using `href=""` may not work on [certain browsers (IE)](https://stackoverflow.com/a/8174417) although it works on all major modern browsers (Chrome, Firefox, Edge, Opera, Safari)

DarrenWestwood, line 57: Changes are working well. It should be noted using `href=""` may not work on [certain browsers (IE)](https://stackoverflow.com/a/8174417) although it works on all major modern browsers (Chrome, Firefox, Edge, Opera, Safari)

DarrenWestwood (approved):

Issue #274, opened 2022-05-23
AngularJS dependency can be removed


blockonomics:
Research the integrations dependency on AngularJS in the checkout. Is this required as mostly we require on some minor JS for async HTTP calls

PR #276, opened 2022-06-23
(update) Remove Angular Dependency, add new standalone PHP Component,…

thisisayush:
This PR references https://github.com/blockonomics/woocommerce-plugin/issues/274

## Updates:
- Standalone PHP Component
- Separate Error Page [Template]
- No Angular-Dependency
- Checkout.js can be hosted in CDN and can be removed from the Repository Directly allowing it to use on Various Platforms
- Checkout.js has no Platform Specific Dependency i.e. No Dependency on any WordPress Feature

## Things Tested and Working:
- Reconnecting Websocket
- QR Code
- Component Rendering via JS
- Support for Following Errors:
  - Display Error
  - BTC Generation Error
  - BCH Generation Error
  - Underpayment Error
  - Payment Expired
- Click to Copy
- Tick (Progress for Timer)
- Expiration of Timer
- Click to Refresh
- Multi-Lingual Support

## Things Not Tested Yet, but supposedly working:
- Support for Following Errors:
  - Gap Limit/Other API Errors
- BCH Payments

## Other Addressed Fixes
- No JS Checkout does not redirect to the error page on Underpayment


templates/blockonomics_error.php
```diff
@@ -0,0 +1,76 @@
...
+<?php
+    } else if ($error_type == "display_error") {
+?>
+    <!-- Display Error -->
+    <div id="display-error" class="bnomics-display-error">
+        <h2><?=__('Display Error', 'blockonomics-bitcoin-payments')?></h2>
+        <p><?=__('Unable to render correctly, Note to Administrator: Please enable lite mode in the Blockonomics plugin.', 'blockonomics-bitcoin-payments')?></p>
+    </div>
+<?php
+    } else if ($error_type == "address_generation_btc") {
```
DarrenWestwood, line 37: We have lots of repeat code to display errors. Let's pass the error title and error message from php to the template to try reserve the template for displaying of the errors. We are already doing this for some errors returned from new address generation.
```
<h2><?php echo $error_title; ?></h2>
<p>
    <?php echo $error_msg; ?>
</p>
```

js/checkout.js
```diff
@@ -0,0 +1,242 @@
...
+
+        //Reload the page if user clicks try again after the order expires
+        this._try_again.addEventListener('click', () => location.reload())
+
+        // Load data attributes
+        let data_container = this.container.querySelector(".blockonomics-data")
+        Object.keys(data_container.dataset).forEach(key => {
+            this.data[key] = data_container.dataset[key]
+        })
+        this.data.time_period = isNaN(this.data.time_period) ? 10 : Number(this.data.time_period)
```
DarrenWestwood, line 135: When collecting the time_period in php we are already setting the default value to 10: https://github.com/blockonomics/woocommerce-plugin/pull/276/files#diff-8d81ebce00651fad5f0a421bc7d60d514e8ca6bd12cc9fc434ab5e46dab827ecR33

js/checkout.js
```diff
@@ -0,0 +1,242 @@
...
+class Blockonomics {
+
+    constructor({
+        checkout_id = 'blockonomics_checkout'
+    }={}) {
+
+        // User Params
+        this.checkout_id = checkout_id
+
+        this.data = {
```
DarrenWestwood, line 11: Defining the expected keys is not necessary. The following will be sufficient:
```
this.checkout_id = checkout_id
this.data = {}
this.progress = {}
```

thisisayush:
@DarrenWestwood @shivaenigma

Updated: 

- Added Checkout Helper and Updated Checkout Templates: This allows the order PHP Logic to be separated from template making templates reusable across platforms and easy for merchants to overwrite, the essential variables are exposed to templates, information about which is available on top of the template page. See `templates/checkout_helper.php`.

- Added Error Helper and Updated Error Templates: The Error Helper is included in the Checkout Helper and works as a middleware to check if there was any error like address generation, underpayments, etc, and loads the error template instead of the checkout template. Merchants can overwrite the error template if they desire. Relevant variables are exposed to the template page for use. Since the template is generic, it can be used on other platforms as well after specific iterations.


Suggested Changes to be Added:
- Move Translation Strings to Helpers from Templates, it allows the template to be truly generic, right now this seems to be the only thing that uses WordPress-related features in templates.
- Unify JS/NoJS Templates and use the JS to dynamically add the required components like Progress Bar, Click to Copy Component (https://github.com/DarrenWestwood/copy-component), and Payment Expired Wrappers. This makes the template customizable as well as the JS Reusable across platforms.

templates/_blockonomics_error_helper.php
```diff
@@ -0,0 +1,41 @@
...
+
+if (isset($order) && $order['status'] == -2) {
+  // Partial payment is recevied
+  $error_type = 'paid_amount_is_less';
+}
+
+$error_title = NULL;
+$has_error = TRUE;
+
+if ($error_type == 'paid_amount_is_less') {
```
blockonomics, line 15: Such complex logic is not required anymore. As I mentioned in #279 we just now need one error message for everything

blockonomics:
We must decrease complexity and not increase it. 
`templates/_blockonomics_error_helper.php` is technically not a template. It has logic etc on what needs to be rendered and is better as part of php/blockonomics.php.

A big problem with php code is the mixing of html code with php logic. This make code unreadable and merchant won't be able to make any changes to template. It is desirable to have a clear distinction. See best practices from wordpress itself as itself has themes / sub templating etc. A suggestion:
- Templates having HTML code that can be changed (having php variables and very less logic)
- Php functional logic is in different files that merchant won't be touching

templates/blockonomics_checkout_helper.php
```diff
@@ -0,0 +1,74 @@
...
+      $template_name = "blockonomics_nojs_checkout";
+    }
+    
+    set_query_var('order', $order);
+    set_query_var('order_id', $order_id);
+    set_query_var('order_amount', $order_amount);
+    set_query_var('crypto', $crypto);
+    set_query_var('payment_uri', $payment_uri);
+    set_query_var('qrcode_url', $qrcode_url);
+
```
blockonomics, line 54: Duplicate and repetitive code, better to avoid/restructure

thisisayush:
@shivaenigma Please review the new changes. Taking your points into consideration, I refactored the load_checkout_template method to autoselect the checkout or error template, and also removed the code causing dual query of Order Details in Checkout Page, the order details are now directly passed to the template code.

Both checkout template and error template now only uses variables and no such PHP Code. 
Checkout Helper is still there because some boilerplate code is needed for JS to work (in the case of JS Template) and because it makes migration for old merchants relatively easy to new templates.

thisisayush:
Referencing #279 Added Generic Customer Facing Error Message instead of Individual Errors. As stated by @shivaenigma, some changes might be required on the Backend to show the GAP Limit Error in the Admin Panel with Test Setup.

image

blockonomics:
I don't see a reason why we should have `checkout_helper` and `error` templates. On seeing the code further, structure is very confusing. `load_checkout_template` does lot of functional logic , a better name would be `process_checkout` which in turn then initializes variables and then loads template 

I also want to discuss about the way we are seeing data-crypto, data-crypto_address etc. variable. I understand this has been done so that JS will be able to access it. However, there is clear duplicate coz we are the setting similar php variables in `checkout_helper` .  The code in `js/checkout.js` works but looks hacky I am unsure of best practices here

thisisayush:
@webmaster

The variables passed to initialize JS was changed to data attributes in this commit: https://github.com/thisisayush/woocommerce-plugin/commit/ed7bca70c52e35ab2d3e8803d146832953bdf103, the reason was to keep the structure as close to previous template to allow easy migration. I guess it's no longer the case.

I agree with checkout helper being an additional file and it can be easily moved to Blockonomics.php but this will require a patch to load_blockonomics_template:

```
// Pseduo Code
...load_blockonomics_template(template_name,...) {

...
// load header

If(template_name_startswith("checkout"))
   Process_checkout_pre_checkout_template()

// Find and load requested template

If(template_name_startswith("checkout"))
   Process_checkout_post_template()

// load footer
...
}
```

thisisayush:
Continuing the discussion about Passing Variables in JS directly or via Data Attributes, we should go for Data Attributes because we can generate the data div in all template files including PrestaShop, WordPress, and WHMCS.

DarrenWestwood:
> assing Variables in JS directly or via Data Attributes, we should go for Data Attributes because we can generate the data div in all template files including PrestaShop, WordPress, and WHMCS.

To confirm, we cannot use the following inline JS as this cannot be done similar in [WHMCS](https://github.com/blockonomics/whmcs-bitcoin-plugin/blob/master/modules/gateways/blockonomics/assets/templates/checkout.tpl#L119) and [Prestashop](https://github.com/blockonomics/prestashop-plugin/blob/master/views/templates/front/payment.tpl)?
```
<script type="text/javascript">
    let blockonomics = new Blockonomics({
  	checkout_id: 'blockonomics_checkout',
  	crypto: '<?php echo $crypto['code']; ?>',
  	order_id: '<?php echo $order_id; ?>',
  	crypto_amount: <?php echo $order['satoshi']; ?>,
  	crypto_address: '<?php echo $order['address']; ?>',
  	fiat_currency: '<?php echo $order['currency']; ?>',
  	fiat_amount: <?php echo $order['value']; ?>,
  	finish_order_url: "<?php echo $blockonomics->get_wc_order_received_url($order_id); ?>",
  	time_period: <?php echo get_option('blockonomics_timeperiod', 10); ?>,
    });
    blockonomics.init();
</script>
```

Please also confirm the following existing code will work with WHMCS and Prestashop or else we should consider removing this from the template file. We can revert to directly running the required checkout JS or moving `let blockonomics = new Blockonomics();` into checkout.js.
```
<script type="text/javascript">
    let blockonomics = new Blockonomics();
    blockonomics.init();
</script>
```

Can we also just call init() from the constructor function?

DarrenWestwood:
> I prefer whatever aligns with best practices and doesn't cause code duplication

We can use the Wordpress [wp_localize_script()](https://developer.wordpress.org/reference/functions/wp_localize_script/) function to pass the PHP variables to JS directly. The same/similar behavior will add overhead when applying to other integrations but can be achieved by looking at how Wordpress enqueue script is working:

- https://stackoverflow.com/a/26639988
- https://github.com/WordPress/wordpress-develop/blob/6.0/src/wp-includes/class.wp-scripts.php#L518-L530

In the background wp_localize_script() prints our PHP variables as inline JS before loading the script:
```
<script> var mydata = {"value":20};</script>
<script src="checkout.js" type="text/javascript"></script>
```

thisisayush:
Tested working of using inline javascript in Prestashop, and WHMCS is already loading JS via inline script tags. IMO initializing the script directly gives us an advantage here. Prestashop only has `register_javascript` hook which can load JS Files, and WHMCS uses direct loading of JS Files.

As for JS, we can use the [wp_localize_script](https://developer.wordpress.org/reference/functions/wp_localize_script/) function or [wp_add_inline_script](https://developer.wordpress.org/reference/functions/wp_add_inline_script/) hook to load the required JS.

To Summarise, **let's move to JS Based Initialization in all platforms**, as every platform supports it. To make things simpler for Merchants to edit the Checkout Template, let's move JS to wp_add_inline_script and initialize the JS from there.

The code below can be changed as per platform while keeping the templates the same (except for TPL/PHP Syntax) which can be resolved in WP by using [Smarty Template Engine ](https://www.smarty.net/crash_course).

```
<!-- The code below is not for Smarty, just pure PHP -->
<script type="text/javascript">
    let blockonomics = new Blockonomics({
  	checkout_id: 'blockonomics_checkout',
  	crypto: '<?php echo $crypto['code']; ?>',
  	order_id: '<?php echo $order_id; ?>',
  	crypto_amount: <?php echo $order['satoshi']; ?>,
  	crypto_address: '<?php echo $order['address']; ?>',
  	fiat_currency: '<?php echo $order['currency']; ?>',
  	fiat_amount: <?php echo $order['value']; ?>,
  	finish_order_url: "<?php echo $blockonomics->get_wc_order_received_url($order_id); ?>",
  	time_period: <?php echo get_option('blockonomics_timeperiod', 10); ?>,
    });
    blockonomics.init();
</script>
```

DarrenWestwood:
> As for JS, we can use the [wp_localize_script](https://developer.wordpress.org/reference/functions/wp_localize_script/) function or [wp_add_inline_script](https://developer.wordpress.org/reference/functions/wp_add_inline_script/) function to load the required JS.

Correct we should use Wordpress methods to pass the PHP variables to our checkout.js. We should use wp_add_inline_script as this is the [best practice](https://developer.wordpress.org/reference/functions/wp_localize_script), however we still only want to parse the PHP array to JS Object to keep things simple. Other integrations will use their relevant methods to perform this if they have.
```
// only pass variables required by checkout.js
wp_add_inline_script( 'blockonomics_checkout', 'const blockonomics_data = ' . json_encode( array(
    'crypto' => $crypto['code'],
    'crypto_amount' => $order['satoshi'],
    ...
) ), 'before' );
```

> The code below can be changed as per platform while keeping the templates the same (except for TPL/PHP Syntax)

The code looks incorrect as we do not want to have any init function printed in the template and WordPress will be handling passing the variables to checkout.js. The correct code on other integrations will look something as follows, as this is how WordPress is currently handling:
```
<script>
const blockonomics_data = {
  crypto: '<?php echo $crypto['code']; ?>',
  crypto_amount: <?php echo $order['satoshi']; ?>,
  ...
};
</script>
<script src="checkout.js" type="text/javascript"></script>
```

checkout.js can then use the data by accessing the blockonomics_data object:
```
const crypto = blockonomics_data.crypto
```

> which can be resolved in WP by using [Smarty Template Engine ](https://www.smarty.net/crash_course).

We should not do this as using Smarty Template Engine in wordpress plugins does not seem to be a Wordpress standard or norm.

thisisayush:
Updated PR. Some known issues I encountered while testing might be present in previous releases as well, but might have gone unnoticed:

1. Display Error Message is triggered by CSS within 3 seconds, this is a problem in Slow Computers or Slow Networks, where CSS is loaded, and while JS processes, the DIsplay Error might be visible. Increasing timeout to 10 seconds compensates for SLow Networks (tested with 50Kb/s) but doesn't compensate for Slow Computers. But seeing that very few or no such issues are reported yet, we can keep it at 10.
2. Script now uses DOMContentLoaded JS Event to trigger the script initialization, this is to ensure that before JS initializes the DOM has been loaded and elements are present. With the earlier scenario, the JS was being initialized after the Elements so it wasn't required, but it presents the issue mentioned above, i.e. DOM may load slowly on some computers which might trigger a Display Error Message, but it is removed as soon as JS is initialized.

DarrenWestwood:
> Display Error Message is triggered by CSS within 3 seconds, this is a problem in Slow Computers or Slow Networks, where CSS is loaded, and while JS processes, the DIsplay Error might be visible. Increasing timeout to 10 seconds compensates for SLow Networks (tested with 50Kb/s) but doesn't compensate for Slow Computers. But seeing that very few or no such issues are reported yet, we can keep it at 10.

This matches the current behavior and should not be changed in this pull request. We can look into this separately if issues are reported. 

__________________________
> Script now uses DOMContentLoaded JS Event to trigger the script initialization, this is to ensure that before JS initializes the DOM has been loaded and elements are present. With the earlier scenario, the JS was being initialized after the Elements so it wasn't required, but it presents the issue mentioned above, i.e. DOM may load slowly on some computers which might trigger a Display Error Message, but it is removed as soon as JS is initialized.

This should not be causing any issues when using `#deferload` as we were previously doing. Please revert and add `#deferload` to wp_enqueue script in order to resolve.

__________________________
```
new Blockonomics().init()
```
Is there a particular reason we are not calling the init function directly from the constructor? https://stackoverflow.com/a/9271352 The constructor can also further be simplified as shown below
```
constructor({
    checkout_id = 'blockonomics_checkout'
}={}) {
    // User Params
    this.checkout_id = checkout_id
    this.init()
}
```

__________________________
```
window.Blockonomics = Blockonomics
```
This is no longer required in checkout.js
__________________________
```
$script = "window.blockonomics_data =
and
this.data = JSON.parse(window.blockonomics_data)
```
It should not be required to attach variables to the global window object as this is generally not a good practice. The following should now be working and aligns with how wordpress handles this in the past using wp_localize_script.
```
$script = "const blockonomics_data =
and
this.data = JSON.parse(blockonomics_data)
```

js/checkout.js
```diff
@@ -0,0 +1,190 @@
...
+        this.progress = {}
+    }
+
+    init() {
+        this.container = document.getElementById(this.checkout_id);
+        if (!this.container) {
+            throw Error(`Blockonomics Initialisation Error: Container #${this.checkout_id} was not found!`)
+        }
+
+        if (!window.blockonomics_data) {
```
DarrenWestwood, line 23: This check should not be required since we also check for a valid JSON object later: https://github.com/blockonomics/woocommerce-plugin/pull/276/files#diff-12c8849e0a66d1a1666bb20d82b2a73b44af2208c029260cfe9db50b35515666R75-R79
The valid JSON check should rather be moved into the init function as it makes more sense to define `this.data = JSON.parse(blockonomics_data)` in init() rather than create_bindings().

thisisayush:
> This matches the current behavior and should not be changed in this pull request. We can look into this separately if issues are reported.

Updated

> This should not be causing any issues when using #deferload as we were previously doing. Please revert and add #deferload to wp_enqueue script in order to resolve.

Deferload triggers JS when the HTML Parsing is completed, which is the same as DOMContentLoaded, moving to defer load won't make a difference here.

> Is there a particular reason we are not calling the init function directly from the constructor? https://stackoverflow.com/a/9271352 The constructor can also further be simplified as shown below

Yes, we can initialize the JS directly with the constructor as well, It was kept like this based on the code before. 

> It should not be required to attach variables to the global window object as this is generally not a good practice. The following should now be working and aligns with how WordPress handled this in the past using wp_localize_script.

Agreed. However, it helps us in 2 Step Debugging to know whether the Script failed because Inline Script failed to load, or because there was some error in JSON. We can remove the window binding and can check it directly inside a try/catch to have the same behavior, or we can remove the 2 Step Check directly, and just focus on having a valid JSON. Using namespaced global variables is used by a lot of plugins/js libraries.

php/Blockonomics.php
```diff
@@ -401,50 +401,58 @@ public function is_lite_mode_active(){
...
+            get_header();
         }
     }
 
     // Adds the footer to the blockonomics page
     public function load_blockonomics_footer($template_name){
         // Lite mode will render without wordpress theme footers
-        if($this->is_lite_mode_active()){
-            // Only load the lite mode javascript if nojs is not active
-            if (!$this->is_nojs_template($template_name)) {
```
DarrenWestwood, line 425: We have unnecessarily moved the lite mode JS from [the bottom footer of the page](https://github.com/blockonomics/woocommerce-plugin/pull/276/files#diff-a3307e680eb76b9d93556c9ae2e5dc134b3526246480730fb8ddf49186da29ebL427) to [the top header of the page](https://github.com/blockonomics/woocommerce-plugin/pull/276/files#diff-a3307e680eb76b9d93556c9ae2e5dc134b3526246480730fb8ddf49186da29ebR420-R423). Please revert this change.

js/checkout.js
thisisayush: Updated based on this.

DarrenWestwood:
> Deferload triggers JS when the HTML Parsing is completed, which is the same as DOMContentLoaded, moving to defer load won't make a difference here.

#deferload is in the existing code and working correctly, which also does not require additional JS code. Please revert back to using #deferload and remove DOMContentLoaded to avoid unnecessary changes.

> Yes, we can initialize the JS directly with the constructor as well, It was kept like this based on the code before.

Yes, let's refactor code as this is no longer the case

> Agreed. However, it helps us in 2 Step Debugging to know whether the Script failed because Inline Script failed to load, or because there was some error in JSON. We can remove the window binding and can check it directly inside a try/catch to have the same behavior, or we can remove the 2 Step Check directly, and just focus on having a valid JSON. Using namespaced global variables is used by a lot of plugins/js libraries.

In the meantime, we can remove the 2 Step Check directly, and just focus on having a valid JSON, as both of these errors will indicate their is an issue with the supplied checkout data. If we have regular issues due to the supplied data, we can consider improving this.

blockonomics-woocommerce.php
```diff
@@ -40,6 +40,9 @@
 require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
 require_once ABSPATH . 'wp-admin/install-helper.php';
 
+// Define constant for Plugin Directory
+define( 'BLOCKONOMICS_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
```
DarrenWestwood, line 44: This is no longer in use and can be removed

js/checkout.js
```diff
@@ -0,0 +1,193 @@
+'use strict';
+class Blockonomics {
+
+    constructor({
```
DarrenWestwood, line 4: The constructor still seems unnecessarily complicated. Please update to the following, unless there is a reason we require the additional overhead
```
constructor({
    checkout_id = 'blockonomics_checkout'
}={}) {
    this.checkout_id = checkout_id
    this.init()
}
```

php/Blockonomics.php
DarrenWestwood: > Moving it to footer will require the additionalScript passed to Footer as well. Why not move all JS to footer?

`wp_add_inline_script('bnomics-checkout', $additional_script, 'before');` and `<script src="<?php echo plugins_url('js/checkout.js', dirname(__FILE__));?>"></script>` should not currently be working together and will require inline usage to fix. Yes we can move loading all JS to the footer if this will be better.

```
<script>
const checkout_data = {
  crypto: '<?php echo $crypto['code']; ?>',
  crypto_amount: <?php echo $order['satoshi']; ?>,
  ...
};
</script>
<script src="checkout.js" type="text/javascript"></script>
```

thisisayush: @DarrenWestwood wp_enqueue_scripts action must be called before calling get_header() in WP. Moving the code to load_blockonomics_footer will require $additional_script to be passed as well to the footer. 

Loading of script in footer is controlled by $in_footer param in [wp_enqueue_script](https://developer.wordpress.org/reference/functions/wp_enqueue_script/) method. 

I've added the [commit for reference](https://github.com/blockonomics/woocommerce-plugin/pull/276/commits/68078d7ab89b8a2146b9e4209211b8641703095d) on how it'll be done.

blockonomics-woocommerce.php
thisisayush: Done

js/checkout.js
thisisayush: Simplified

thisisayush:
> In the meantime, we can remove the 2 Step Check directly, and just focus on having a valid JSON, as both of these errors will indicate there is an issue with the supplied checkout data. If we have regular issues due to the supplied data, we can consider improving this.

I added the [fallback](https://github.com/blockonomics/woocommerce-plugin/pull/276/commits/db6956dfd0f7195ae45c8c91d3145e0f73cf8390#diff-12c8849e0a66d1a1666bb20d82b2a73b44af2208c029260cfe9db50b35515666R26-R30) to know if it was a JSON Issue or a Reference issue.

thisisayush:
> #deferload is in the existing code and working correctly, which also does not require additional JS code. Please revert back to using #deferload and remove DOMContentLoaded to avoid unnecessary changes.

Moving the scripts to the footer no longer requires this, since DOM Is available by then

php/Blockonomics.php
DarrenWestwood: Loading scripts in the head and using defer is considered best for performance as the script is loaded asynchronously and only executed after HTML parsing: https://flaviocopes.com/javascript-async-defer/#with-defer-in-the-head

Using defer in the plugin was initially introduced to remove the "Eliminate render-blocking resources" warnings from lighthouse.
While loading the script in the footer is also suitable as it does not throw this warning, the performance may be slightly affected and will require further testing. It will be simpler to use defer as this is existing method and most efficient.

Lets move the lite mode JS back into the header and add the defer attribute to keep the lite mode consistent with the standard checkout. We can then only pass additional_script to the header again as well.

thisisayush: Okay

thisisayush:
Updated

php/Blockonomics.php
```diff
@@ -401,50 +401,64 @@ public function is_lite_mode_active(){
...
-        if (!$this->is_nojs_template($template_name)) {
-            add_action('wp_enqueue_scripts', 'bnomics_enqueue_scripts' );
-        }
+    public function load_blockonomics_header($template_name, $additional_script=NULL){
         // Lite mode will render without wordpress theme headers
         if($this->is_lite_mode_active()){
         ?>
-          <link rel="stylesheet" type="text/css" href="<?php echo plugins_url('css/order.css', dirname(__FILE__));?>">
+            <link rel="stylesheet" type="text/css" href="<?php echo plugins_url('css/order.css', dirname(__FILE__));?>">
+            <script src="<?php echo plugins_url('js/vendors/reconnecting-websocket.min.js', dirname(__FILE__));?>" defer="defer"></script>
```
DarrenWestwood, line 409: Both light mode and nojs can be activated at the same time. We should still be checking `if (!$this->is_nojs_template($template_name))` when light mode is active. Currently the JS is loaded when both lite mode and nojs settings are activated.

thisisayush:
Updated

DarrenWestwood (approved):
The changes are looking good and functionality has been tested to be working correctly.

PR #285, opened 2022-08-26
(fix) Error Template should not load JS

### LINKED ISSUE #284: Console Error blockonomics_checkout div not found
This issue was found during testing after changes in https://github.com/blockonomics/woocommerce-plugin/pull/276

image

thisisayush:
Closes #284

### DIFF
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -400,17 +400,25 @@ class Blockonomics
         return get_option('blockonomics_lite', false);
     }
 
+    public function is_error_template($template_name) {
+        if (strpos($template_name, 'error') === 0) {
+            return true;
+        }
+        return false;
+    }
+
     // Adds the header to the blockonomics page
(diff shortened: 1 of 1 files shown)


thisisayush:
@DarrenWestwood This case was left and was identified by @shivaenigma and @ankit61d, the issue lies in the fact that JS is set to auto initialize, so it should not be loaded when an error occurs because required context information and required DOM Structure isn't available there.

The fix works by adding a check for detecting the error template and ignoring the JS to load as we do in No JS Templates.

DarrenWestwood (approved):
Approving as these changes are a working fix for the issue. These changes can be merged, in order to be included in the upcoming code refactor. 

Refactor Note: A single check if the template is the js checkout template, rather than comparing against all other templates `if (!$is_nojs_template && !$is_error_template)` makes more sense when checking to load the javascript for the checkout

PR #287, opened 2022-09-02
(fix) generate QR Code without GD extension

### LINKED ISSUE #286: Make QR code generation work without GD extension
`PHP Fatal error:  Uncaught Error: Call to undefined function ImageCreate() in /home/enigma/blockonomics-bitcoin-payments/php/phpqrcode.php:999`

https://stackoverflow.com/questions/3106991/fatal-error-call-to-undefined-function-imagecreate
We don't expect merchant to have GD extension as this is not a requirement for wordpress

ankit61d:
This fixes #286 by using [this library](https://github.com/splitbrain/php-qrcode/blob/master/src/QRCode.php) to generate QR Code as a svg element. This renders QR code when NoJS mode is active and  it removes need to have GD extension enabled.


DarrenWestwood:
PR is working to fix the GD extension error. It should be noted that simply changing `QRcode::png($codeText);` to `QRcode::svg($codeText);` also solves the issue without changing libraries. Please select the better of the 2 libraries based on how recently the library is updated and the PHP compatibility of the library.

The qr code sizing and spacing also differs between js and nojs versions. The centering of the 'Open in wallet' link below the qrcode is slightly off with the nojs qr. We should look into better improving the consistency between the two versions if this can be done without too many changes.
JS:
JS
No JS:
NoJS

ankit61d:
Closing this PR, this issue is now addressed in #288 in https://github.com/blockonomics/woocommerce-plugin/pull/288/commits/0669e8ebfd63f04f36bb29a6776f6dfc8e538c45

Issue #282, opened 2022-08-26
Underpayment - Allow customer to pay more


blockonomics:
Currently on underpayment customer has no option than to wait for merchant. He can be given an invoice to pay the pending amount. Also we have think if underpayment slack will still work

blockonomics:
Check woocommerce order actions . Merchant has option to accept order or send invoice to customer for more payment

blockonomics:
The idea here is the following. On underpayment, the merchant should get following options to do on opening Woocommerce > Orders > Order Details

1. Show the received USD amount vs Order Amount clearly to the merchant. Eg 18USD was received for 20USD other
2. Merchant has two options:
   i. Accept the order
   ii. Invoice the customer for remaining 2USD 
3. Also indicate underpayment slack option to merchant avoid future problems

ankit61d:
Sounds an interesting problem to work upon.

ankit61d:
Order details after underpayment for an order (Merchant setting has 0% underpayment slack). 
Note that (i) it is marked as failed so it is no longer editable, and (ii) merchant can send "Note to Customer".

![2209-000792](https://user-images.githubusercontent.com/97018228/191675423-f21b72a5-08ea-42fb-9591-7e9b5d8be0df.jpg)




> Show the received USD amount vs Order Amount clearly to the merchant

We can add 2 more fields for this as seen here:
![2209-000793](https://user-images.githubusercontent.com/97018228/191672517-8494f53b-ea08-4574-8060-5ea96989ee42.jpg)

> i. Accept the order


For this merchant only need update the order status to "Completed":
<img width="413" alt="image" src="https://user-images.githubusercontent.com/97018228/191672909-d8dd70df-f012-4b93-8fa7-a2bbbe266fa7.png">

> ii. Invoice the customer for remaining 2USD. On receiving this we have to mark invoice paid

The first thing merchant has to do is change the status back to "Pending Payment" so as to make the order editable again. Now merchant can email invoice order details from order action. 
![2209-000800](https://user-images.githubusercontent.com/97018228/191682546-e6eadbdd-5b02-4841-8887-a4873f70ed7a.jpg)

Here is the email customer receives:
<img width="413" alt="image" src="https://user-images.githubusercontent.com/97018228/191681539-cdf39097-fe42-4626-8d62-6eb8847880b8.png">

 The link in email will generate this payment page for customer:
[...]

PR #291, opened 2022-10-10
Coupon implementation on underpayment

ankit61d:
**#282 Phase 1**  - as discussed [here](https://github.com/blockonomics/woocommerce-plugin/issues/282#issuecomment-1262360818) on issue #282 

Add automatically generated coupon on confirmed underpaid callback. 
Discount amount is calculated in 68c8396

image

Coupon menu in merchant dashboard
image

Trying to add same coupon on another checkout throws usage limit error:

Now, if the order status is set to failed or cancelled, this coupon's usage/limit changes to 0/1, so it can be used one more time. If we want to prevent that, there are 2 ways:
1. We can set auto-expiry to 1 hour (time after which woocommerce itself cancels the order), or 
2. We can write an action hook that will see if the order failed/cancelled had a coupon code starting with 'bck_' with total length = 10. - done in https://github.com/blockonomics/woocommerce-plugin/pull/291/commits/1f63c70c829088a1c6e2d01d79f109c231830af9 -- reverted in 04d696c as per discussion below.

-----
Customer now sees the order with discount applied
image


blockonomics-woocommerce.php
```diff
@@ -58,6 +58,8 @@ function blockonomics_woocommerce_init()
     add_action('woocommerce_email_customer_details', 'nolo_bnomics_woocommerce_email_customer_details', 10, 1);
     add_action('admin_enqueue_scripts', 'blockonomics_load_admin_scripts' );
     add_action('restrict_manage_posts', 'filter_orders' , 20 );
+    add_action('woocommerce_order_status_failed','delete_underpayment_coupon');
+    add_action('woocommerce_order_status_cancelled', 'delete_underpayment_coupon');
```
shivaenigma, line 62: This is a bit hacky. Lets not do this right now

shivaenigma:
>Now, if the order status is set to failed or cancelled, this coupon's usage/limit changes to 0/1
This is not a big deal really.  Few things:

- Does the customer ever even see the coupon code to be able to use the exploit , no I guess ! I am not seeing it anywhere in the customer UI/emails
- Logically if the order got cancelled, the customer is atleast entitled to the coupon for the BTC he spent

ankit61d:
**Update on Phase 2**:
I have added a new field in blockonomics order table - _timestamp_.  For a new order, timestamp records the time of order creation. Now, after confirmed underpaid callbacks:

1.  Coupon is applied to the order equal to amount received in the callback
2. After applying coupon, new address is generated and a new row is added in the blockonomics order table. So on each underpaid callback the

blockonomics-woocommerce.php
```diff
@@ -501,10 +501,16 @@ function blockonomics_activation_hook() {
...
+// Adding timestamp column in db to handle underpayments
 function blockonomics_update_db_check() {
     global $wpdb;
     global $blockonomics_db_version;
 
+    $timestamp_column = $wpdb->get_results(  "SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS
+    WHERE table_name = 'wp_blockonomics_orders' AND column_name = 'timestamp'"  );
+    if(empty($timestamp_column)){
+        $wpdb->query("ALTER TABLE wp_blockonomics_orders ADD timestamp INT DEFAULT NULL");
+    }
```
blockonomics, line 513: Be very careful what you put in the blockonomics_update_db_check() function. It is loaded for every PHP page by `plugins_loaded` . Add a doc comment on top of function for this
The upgrade code must be inside  below if statement

blockonomics-woocommerce.php
```diff
@@ -501,10 +501,16 @@ function blockonomics_activation_hook() {
...
     $installed_ver = get_site_option( 'blockonomics_db_version' );
     if ( $installed_ver != $blockonomics_db_version ) {
```
blockonomics, line 515: Upgrade should happen here only trigger the migration code related to upgrade. I am not sure why we are running blockonomics_create_table() for everything. @DarrenWestwood  can you elaborate on this

Another thing to think about is that existing orders will have timestamp NULL. We have to think through implications of this if any. If underpayment doesn't work on existing orders its ok, but ongoing callbacks to existing orders should not get effected

php/Blockonomics.php
```diff
@@ -521,7 +521,8 @@ public function create_new_order($order_id, $crypto){
                 'order_id'           => $order_id,
                 'status'             => -1,
                 'crypto'             => $crypto,
-                'address'            => $address
+                'address'            => $address,
+                'timestamp'          => time()
         );
```
blockonomics, line 526: I understand the change here and how it works. However logically it feels weird as same order_id has multiple status and we really don't know whether order has been paid or unpaid

A better structure would be an order and a payments table
Again @DarrenWestwood  feel free to let us know your thoughts

blockonomics-woocommerce.php
DarrenWestwood: These upgrade changes should be moved into their own db version if statement and the database version incremented: `$blockonomics_db_version = '1.2';`
```
if ($installed_ver < 1.2) {
  // Upgrade code
}
```

Using the `plugins_loaded` hook is the recommended method for running the upgrade to ensure this runs on plugin updates. More information [here](https://codex.wordpress.org/Creating_Tables_with_Plugins#Adding_an_Upgrade_Function).

blockonomics-woocommerce.php
DarrenWestwood: > Upgrade should happen here only trigger the migration code related to upgrade. I am not sure why we are running blockonomics_create_table() for everything.

Yes blockonomics_create_table should only be run if there is no $installed_ver. I have opened a separate issue to deal with this:  https://github.com/blockonomics/woocommerce-plugin/issues/296

> Another thing to think about is that existing orders will have timestamp NULL

Adding timestamp may not be necessary as we should be able to simply check for pending address: https://github.com/blockonomics/whmcs-bitcoin-plugin/blob/master/modules/gateways/blockonomics/blockonomics.php#L426-L436 (There can only be one pending address at a time for each crypto, therefore it is not necessary to order addresses by timestamp to locate the pending address)

php/Blockonomics.php
DarrenWestwood: > logically it feels weird as same order_id has multiple status

The table name `blockonomics_orders` is causing this confusion. This table should be viewed as a list of `crypto_addresses` generated by Blockonomics and their relevant data. The order_id therefore tells us which woocommerce order the blockonomics address is linked to and the status for the address.

> we really don't know whether order has been paid or unpaid

In the past we were tracking the expected `satoshi` in the db and `paid_<crypto>` amount in the post meta (custom fields). The `paid_<crypto>` amount in the post meta has been adapted in this PR to `paid_order_amount` (fiat) and `paid_to_<address>` (btc). This seems to be functioning well, however satoshi should rather be used or we should save the `satoshi_payed` to the database, and continue displaying btc/fiat in the post meta.

blockonomics: Yes I undestand it is crypto_addresses but to the simple question, was the order paid or underpaid is not payed is difficult in the current structure.
The two table structure keeps order payments into a separate table and  is also directly inspired from WHMCS

blockonomics: Feel free to create a document with comparsion of the two structures

DarrenWestwood: The `satoshi_payed` column should make it easier to see if each address received the correct amount and matches the same structure from WHMCS:
https://docs.google.com/document/d/18aoTcTA6iPERgCGYPtHQI4o8GqutmXS4BYsuaGT8Kgg/edit?usp=sharing

ankit61d:
3 possible changes to DB discussed here:
https://docs.google.com/document/d/1m2lX_ZMCupZHDvkjdauBEkb4fGhA5EWXTrvflDjuhhg/edit?usp=sharing

ankit61d:
Update: https://github.com/blockonomics/woocommerce-plugin/pull/291/commits/09db6e7cbb3d1e57d79ef4a6b9fe05f534bfb09c - Now customer automatically receives email on confirmed underpayment, with a link to pay for remaining amount.
Email screenshot:
image

https://user-images.githubusercontent.com/97018228/202644241-e46f0d1c-b24b-4624-b876-3eeeba3eea1f.mp4

DarrenWestwood (commented):
Coupon discounts for underpayments are working well. Some functionality appears to be happening during callbacks, which can possibly be moved into the order processing functionality where required.

shivaenigma:
Not implementing . Closing in favour of https://github.com/blockonomics/woocommerce-plugin/pull/335

Issue #283, opened 2022-08-26
Upgrade to more modular Checkout UX 


blockonomics:
![Property_1payment](https://user-images.githubusercontent.com/22165583/186842256-3279ebae-c7bd-400b-b261-17fbcffb4b30.png)
### Desired Behaviour/Motivation
- Timeout is confusing and it seems to customer that if he doesn't send within that time, his payment is lost
- If time reaches 2 min or so, customer is basically left with nothing to do as he knows he cannot complete payment in that time
- A modular UX clearly indicates that the timer is for price refresh and address stay the same

blockonomics:
Lets start working on this for wordpress:
- Important thing is that we should respect the theme and not do unnecessary styling
- We have to think on how to implement boxes that gel with the theme (We cannot use background color)
- Mobile friendliness is important
- Copy we can the COPY component UX

**Todo for next version**
- Refresh now should only reload the BTC amount section instead of refreshing the entire page. Important that this should not be over-complicated

blockonomics:
>Refresh now should only reload the BTC amount section instead of refreshing the entire page. Important that this should not be over-complicated
@thisisayush Lets think about this while you are waiting for design

thisisayush:
I think this can be achieved by what we had previously, an AJAX API that can be used to refresh the order parameters and show it here. It'll not be possible in NoJS Solution, but we can implement auto page reload in NoJS Scenario as well (using HTML meta page refresh)

blockonomics:
Yes in noJS we can just reload the page. That is not a problem
Lets see what minimalistic AJAX API we can use that other wordpress plugins are using using. Maybe jquery can be used as we know wordpress will have jquery

PR #295, opened 2022-10-23
Phase 1: Upgrade JS Checkout to Modular (Part 2/2)

thisisayush:
References #283

Continues #293, To be Merged after

# Part 2 of JS Modular Checkout:
- Adds support for Ajax Based Refresh 

Working Demo:

screen-recorder-sun-oct-23-2022-13-53-51


thisisayush:
Update: Rebased Branch from the Forked Branch (thisisayush/modular-checkout) [ #293 ], Added Support for Auto Refresh on Order Expire and Fixed Decimal Places

screen-recorder-tue-nov-01-2022-10-08-59

js/checkout.js
```diff
@@ -180,6 +200,76 @@ class Blockonomics {
...
+            this._refresh.querySelector('.blockonomics-icon-refresh').classList.add('spin')
+            this._refresh.setAttribute('disabled', 'disabled')
+        } else {
+            this._refresh.querySelector('.blockonomics-icon-refresh').classList.remove('spin')
+            this._refresh.removeAttribute('disabled')
+        }
+    }
+
+    refresh_order() {
+        let url = `${this.data.api_url}?get_order=${this.get_url_param('show_order')}&crypto=${this.get_url_param('crypto')}`
```
DarrenWestwood, line 215: Please see https://github.com/blockonomics/woocommerce-plugin/pull/276/files#diff-32cdce28ac549728f6a28999f6d874aa4f29e6da1406c28c3134cd8d421a866aL183 for forming the woocommerce url: https://github.com/blockonomics/woocommerce-plugin/blob/3302c63a7faa1346cc1faf6d0671e14a0b86e75e/js/app.js#L183-L200

php/Blockonomics.php
```diff
@@ -631,6 +633,10 @@ public function get_wc_order_received_url($order_id){
         return $wc_order->get_checkout_order_received_url();
     }
 
+    public function get_wc_api_url(){
+        return site_url() . "/wc-api/WC_Gateway_Blockonomics";
```
DarrenWestwood, line 637: Please refer to https://github.com/thisisayush/woocommerce-plugin/blob/435854c3589ab2abece71d097f782f343eaf2cd6/php/Blockonomics.php#L361

js/checkout.js
```diff
@@ -180,6 +200,76 @@ class Blockonomics {
...
+        
+        fetch(url, {method: 'GET'}).then(
+            res => {
+                this._set_refresh_loading(false)
+                if (res.status == 200) {
+                    res.json().then(data => {
+                        this._update_order_params(data)
+                    },
+                    () => {
+                        this._fallback_refresh_order()
```
DarrenWestwood, line 230: Let's remove the fallback functions from this PR as this adds overhead

php/Blockonomics.php
```diff
@@ -572,6 +572,7 @@ public function get_checkout_context($order, $crypto){
                 // Display Checkout Page
                 $context['order_amount'] = $this->fix_displaying_small_values($order['satoshi']);
                 $context['payment_uri'] = $context['crypto']['uri'] . ":" . $order['address'] . "?amount=" . $context['order_amount'];
+                $context['crypto_rate'] = number_format($order['value']/$context['order_amount'], 2, '.', '');
```
DarrenWestwood, line 575: Please add code comment to note we are re-calculating the crypto_rate, which may slightly differ to rate returned by Blockonomics. Please also add reasoning to this PR as to why we are doing this, as opposed to passing this returned value with the $order.

js/checkout.js
thisisayush: Let's keep the one where there is network error, it helps the plugin fallback to legacy mode in case of Issues like CORS.

php/Blockonomics.php
thisisayush: Done

js/checkout.js
thisisayush: Done

js/checkout.js
```diff
@@ -180,6 +200,70 @@ class Blockonomics {
...
+        location.reload()
+    }
+
+    _satoshi_to_amount(satoshi) {
+        return satoshi/1.0e8
+    }
+
+    _price_per_crypto(crypto, fiat_amount) {
+        return Number(fiat_amount/crypto).toFixed(2) // Show Fixed Decimal Points rather than rounding off
+    }
```
blockonomics, line 261: One line functions are a bad practice particularly particularly when the function itself is called from only one place
https://mytwoocents.com/2017/03/25/one-line-methods-can-easily-become-an-antipattern/
In C we have macros for the this same reason

thisisayush: It makes sense, I've updated the PR for the same. I haven't yet replaced `redirect_to_finish_order` method as it was not part of this PR.

thisisayush:
@DarrenWestwood 

As discussed, I've separated the `get_order` endpoint which only contains the relevant information required for Updating in JS, it provided us with the flexibility of using the backend for some things we were calculating in the front, such as the crypto rate. and thus I've moved these things to the API to provide better consistency in data.

Also, I meed a crucial part that was updating QR Code and Payment URIs, which I've added in the latest commit.

PR #298, opened 2022-11-09
Phase 2: Upgrade No JS Checkout to Modular 

thisisayush:
References #283 
Closed in favor of #306

# No JS Modular Checkout:
- Upgrades No JS Checkout to Modular

### Lite Mode
image

### Normal Mode
image


thisisayush:
Quick Question @DarrenWestwood  @shivaenigma , we don't need updates in XX:XX Minutes in this because it doesn't make sense, what should we change it to or should we remove it?

PR #301, opened 2022-11-16
Modular checkout full (Part 1 and Part 2)

DarrenWestwood:
Merged modular checkout changes:
Part 1 (Checkout Display)- https://github.com/blockonomics/woocommerce-plugin/pull/293
Part 2 (Refresh Functionality)- https://github.com/blockonomics/woocommerce-plugin/pull/295


DarrenWestwood:
Pull request is ready for final testing. I am experiencing issues with the copy to clipboard functionality, however this only appears to be on my end. Please verify this is working correctly during testing.

js/checkout.js
```diff
@@ -180,6 +203,67 @@ class Blockonomics {
...
+        clearInterval(this.progress.interval)
+        
+        fetch(this.data.get_order_amount_url, {method: 'GET'}).then(
+            res => {
+                // Enable the button anyways so that user can retry
+                this._set_refresh_loading(false)
+
+                if (!res.ok) {
+                    // Show Error, do not log to console as it is a Server Issue like 404
+                    this.toggle_ajax_error(true)
```
blockonomics, line 236: I think whole toggle_ajax_error() is adding more complexity. Just reload the page if we are having ajax error

blockonomics:
[Screencast from 16-11-22 06:07:14 PM IST.webm](https://user-images.githubusercontent.com/22165583/202187316-d9d611bd-c9c4-4a08-8343-4e10ab03be27.webm)

- Minor problem:  Left align the boxes of each section (not very important) if this adds complexity
- Major problem: I can't notice the refresh unless I keep watching the numbers change. We need to an animation to refresh the whole amount section

php/Blockonomics.php
```diff
@@ -571,7 +581,8 @@ public function get_checkout_context($order, $crypto){
             } else {
                 // Display Checkout Page
                 $context['order_amount'] = $this->fix_displaying_small_values($order['satoshi']);
-                $context['payment_uri'] = $context['crypto']['uri'] . ":" . $order['address'] . "?amount=" . $context['order_amount'];
+                $context['payment_uri'] = $this->get_crypto_payment_uri($context['crypto'], $order['address'], $context['order_amount']);
+                $context['crypto_rate'] = $this->get_crypto_rate_from_params($order['value'], $context['order_amount']);
```
blockonomics, line 585: Formatted numbers are strings so it better to change their variable name so that we are not confused 
['crypto_rate'] can be renamed to ['crypto_rate_str']
['order_amount'] can be renamed to ['order_btc_str']

blockonomics: Also I think $this->get_crypto_rate_from_params should use $order['satoshi'] rather than $order['order_amount'] as we should calculate from the accurate satoshi value rather than the rounded string value

php/Blockonomics.php
```diff
@@ -702,6 +716,24 @@ public function get_order_info($order_id, $crypto){
...
+    public function get_order_amount_info($order_id, $crypto){
+        $order = $this->process_order($order_id, $crypto);
+
+        $order_amount = $this->fix_displaying_small_values($order['satoshi']);
+        
+        $cryptos = $this->getActiveCurrencies();
+        $crypto_obj = $cryptos[$crypto];
+
+        $response = array(
+            "payment_uri" => $this->get_crypto_payment_uri($crypto_obj, $order['address'], $order_amount),
```
blockonomics, line 729: My first reaction is both order_amount and payment_uri can calculated from JS side easily avoiding us to pass extra parameters from server.

It looks like this was done for compatibility with noJS template ? This is fine. We just need to add code comment so that we can understand the motivation of the redundant params being passed from server

js/checkout.js
thisisayush: That was the previous behavior, but Darren suggested that we show the errors to the User as it helps in debugging from the site owner's perspective and to see what's wrong or if anything is amiss. If we reload, it'll never be caught or reported.

blockonomics: Understood. I don't expect lots of errors as everybody supports ajax. Just reloading is easy

DarrenWestwood (commented):
Overall the code looks good and is functioning well. There are some minor changes that still need to be addressed and some minor issues with the refresh price animation.

css/order.css
```diff
@@ -258,9 +250,166 @@ svg {
...
+
+.blockonomics-icon-qr::before {
+  content: "\e901";
+}
+
+.blockonomics-icon-copy::before {
+  content: "\e902";
+}
+
+.blockonomics-icon-refresh::before {
```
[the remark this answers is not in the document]
thisisayush: Yes, updated this, I had to convert the icon downloaded from figma from Strokes to Fills because icomoon wasn't supporting it directly, it's updated now. 

For future reference, if you see this: 
image

Use this website https://iconly.io/tools/svg-convert-stroke-to-fill to convert the icon and then use it in IcoMoon, the documentation in icomoon asks you to do in Illustrator and other professional tools, but if it's a simple icon use this!

js/checkout.js
```diff
@@ -180,6 +203,146 @@ class Blockonomics {
...
+        this._set_refresh_loading(true)
+        this.toggle_ajax_error(false)
+
+        // Stop Progress Counter
+        clearInterval(this.progress.interval)
+        
+        fetch(this.data.get_order_amount_url, {method: 'GET'}).then(
+            res => {
+                // Enable the button anyways so that user can retry
+                this._set_refresh_loading(false)
```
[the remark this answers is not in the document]
thisisayush: Resolved

thisisayush:
@DarrenWestwood All mentioned issues resolved

thisisayush:
Quick Question, why are we not releasing #298 with same release? @DarrenWestwood @blockonomics

DarrenWestwood:
The refresh price animation is currently not working with the Twenty Twenty-Two theme.

DarrenWestwood:
> Quick Question, why are we not releasing https://github.com/blockonomics/woocommerce-plugin/pull/298 with same release?

This PR has been open for long now and we should focus on getting this released. NoJS seems unaffected by the current changes and should not require many changes once this PR is merged. We can look into quickly merging https://github.com/blockonomics/woocommerce-plugin/pull/298 as soon as this PR is done.

thisisayush:
> The refresh price animation is currently not working with the Twenty Twenty-Two theme.

Check the issue, the problem was because of Flex Aligning of the Input Box, during the resolution, another issue that caused sizing to be disrupted was also fixed.

php/Blockonomics.php
```diff
@@ -547,6 +547,16 @@ public function fix_displaying_small_values($satoshi){
         }
     }
 
+    public function get_crypto_rate_from_params($value, $satoshi) {
+        // Crypto Rate is re-calculated here and may slightly differ from the rate provided by Blockonomics
+        // This is required to be recalculated as the rate is not stored anywhere in $order, only the converted satoshi amount is.
```
shivaenigma, line 552: The comment is misleading.
The main requirement of this function is to avoid scientific notification. Crypto rate can be calculated in JS also using satoshi/amount*1.0e8. However, this leads to sometimes scientific notation display like 2.0E107 .

TO avoid this server return a convert string can be be used for display both in JS / nojS templates

thisisayush: This wasn't the motivation towards building this function, the same logic can be implemented in JS without showing the Scientific Notification by formatting number as we do here. The reason this method was required was:

1. We need to show the rate in both JS and No-JS that's why it makes sense to have it in the backend just like the payment_uri (https://github.com/blockonomics/woocommerce-plugin/pull/301/files/b20ff4224e5227ecf7de231838f56834b74897a9#diff-a3307e680eb76b9d93556c9ae2e5dc134b3526246480730fb8ddf49186da29ebR584-R586)

2. The primary concern for this method was the comment that was added and that was the difference in rates that Blockonomics Provide and the rate which is calculated here, and why it is being recalculated. to avoid confusion in the future.

But yes, we can add a comment mentioning why we added it in backend and not frontend directly,

image

DarrenWestwood:
Code looks good and checkout (with refresh animation) is working well with different themes.

PR #303, opened 2022-12-27
Refresh animation timing improvements

DarrenWestwood:
- Refresh time remaining timing delay fix
- Use 300ms for completing animations

### DIFF
--- a/js/checkout.js
+++ b/js/checkout.js
@@ -244,7 +244,7 @@ class Blockonomics {
         let left = parseFloat(style.left.replace("px", ""))
 
         setTimeout(() => {ele.style.left = (width + left) + 'px', ele.style.width = '0px'}, 10)
-        setTimeout(() => ele.remove(), 400)
+        setTimeout(() => ele.remove(), 300)
     }
 
     _animate_price_update() {
@@ -270,7 +270,7 @@ class Blockonomics {
             container.style.position = null
             parent_container.style.overflow = parent_container.getAttribute('data-bnomics-overflow')
             parent_container.removeAttribute('data-bnomics-overflow')
-        }, 400)
+        }, 300)
     }
 
     _set_refresh_loading(loading=false) {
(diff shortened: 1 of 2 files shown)


js/checkout.js
```diff
@@ -102,7 +102,7 @@ class Blockonomics {
             clock: this.data.time_period * 60,
             percent: 100
         }
-
+        this._time_left.innerHTML = `${String(Math.floor(this.progress.clock/60)).padStart(2, "0")}:${String(this.progress.clock%60).padStart(2, "0")} min`
```
thisisayush, line 105: Why not we just do?
```javascript
this.progress.clock += 1
this.tick()
```

js/checkout.js
```diff
@@ -244,7 +244,7 @@ class Blockonomics {
         let left = parseFloat(style.left.replace("px", ""))
 
         setTimeout(() => {ele.style.left = (width + left) + 'px', ele.style.width = '0px'}, 10)
-        setTimeout(() => ele.remove(), 400)
+        setTimeout(() => ele.remove(), 300)
```
thisisayush, line 247: I added a 100ms extra time so that the animation eases out before we remove it, to make it look less abrupt.

js/checkout.js
```diff
@@ -270,7 +270,7 @@ class Blockonomics {
             container.style.position = null
             parent_container.style.overflow = parent_container.getAttribute('data-bnomics-overflow')
             parent_container.removeAttribute('data-bnomics-overflow')
-        }, 400)
+        }, 300)
```
thisisayush, line 273: I added a 100ms extra time so that the animation eases out before we remove it, to make it look less abrupt.

js/checkout.js
```diff
@@ -301,7 +301,7 @@ class Blockonomics {
             }
         ).then(res => {
             this._update_order_params(res)
-            setTimeout( () => this._set_refresh_loading(false), 500)
```
thisisayush, line 304: I added the delay to counter the effects of `reset_progress` (without the change that was done in this PR), so that when the counter resets, it gives us the same time to reset the animations as before the next `tick` is triggered

thisisayush:
Added fix for Refresh Oscillation @DarrenWestwood

PR #306, opened 2022-12-28
Phase 2: Upgrade No JS Checkout to Modular #298

### LINKED ISSUE #305: NOJS modular checkout
image

Currently its showing the above. Which is quite ok. With NOJS we this current limitation:

- We cannot refresh the price after 10 min, so basically customer can pay on same price even after days (kind of small loophole)

Lets discuss what exactly what we propose to do in NoJS

thisisayush:
References #283 
Closes #305 
Continuation for #298 

# No JS Modular Checkout:
- Upgrades No JS Checkout to Modular

### Lite Mode
image

### Normal Mode
image

### DIFF
--- a/css/order.css
+++ b/css/order.css
@@ -1,11 +1,5 @@
 /* ----- Checkout Page Styles ------*/
 
-.no-js .bnomics-order-container {
-  margin: 5vh auto;
-  padding: 8px;
-  max-width: 700px;
-}
-
 .bnomics-order-info {
   display: flex;
 }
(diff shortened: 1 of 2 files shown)


shivaenigma:
Don't do anything in frontend. This is ok

shivaenigma:
No refresh functionality

thisisayush:
Related Issues/PR:
#307 
#309 

The above issues do not affect the current PR in any way, they are additional functionalities/bugs which should be addressed as a separate issue.

DarrenWestwood (approved):

PR #310, opened 2023-02-03
Updated db and underpayment flow

ankit61d:
Picks up coupon implementation on underpayment from https://github.com/blockonomics/woocommerce-plugin/pull/291
And subsequent changes discussed on [Notion](https://www.notion.so/blockonomics/Wordpress-3-5-8-Release-Testing-70a4e71f419d46b39877caa38e4fbe26)

---
**Database Changes**:

Here is the old db schema
```
Table blockonomics_orders {
        order_id int NOT NULL,
        status int NOT NULL,
        crypto varchar(3) NOT NULL,
        address varchar(191) NOT NULL,
        satoshi int,
        currency varchar(3),
        value longtext,
        txid text,
        PRIMARY KEY (address),
        KEY orderkey (order_id,crypto)
  }
```
The changes introduced are:

1. Table name is changed to `blockonomics_payments`
2. `status` is renamed to `payment_status`
3. `satoshi` is renamed to `expected_satoshi`
4. `value` is renamed to `expected_fiat` & it is no longer `longtext`
5. Two new fields are added for logging purpose only - `paid_satoshi` & `paid_fiat` 

The new db schema looks this
```
Table blockonomics_payments {
         order_id int NOT NULL,
            payment_status int NOT NULL,
            crypto varchar(3) NOT NULL,
            address varchar(191) NOT NULL,
            expected_satoshi int,
            expected_fiat double,
            currency varchar(3),
            paid_satoshi int,
            paid_fiat double,
            txid text,
            PRIMARY KEY (address),
            KEY orderkey (order_id,crypto)
        }
```
---

**Payment Status** are now only the following:
- 0 : Meaning payment not started
- 1: Meaning payment in process (unconfirmed payment)
- 2: Meaning payment completed (doesn’t mean order has been paid)

---

**Custom Fields:**
1. Removed `expected_btc` 
2. `paid_btc` is now a hidden field
3. `btc_address` is now `blockonomics_payment_addresses` and contains all (btc or bch) address generated for the order
4. `btc_txid` is now `blockonomics_payments_txid` and includes all (btc or bch) txid for the order
image

---

**Sample email invoice after underpayment**

image

---

Tests Checked:
1. Mix payment with btc / bch ✔
2. Generate new address only after initial payment confirmation ✔
3. order address / txid searchable in single payment ✔
4. order address / txid searchable in multiple payments ❌
5. Email invoice with link to pay ✔


shivaenigma:
>4.order address / txid searchable in multiple payments

Isn't necessary for now as we don't expect also of multiple payment orders

php/Blockonomics.php
```diff
@@ -753,39 +809,60 @@ public function check_callback_secret($secret){
...
+          $order['payment_status'] = 2;
+          $order = $this->check_paid_amount($status, $value, $order, $wc_order);
           $this->update_temp_draw_amount($value);
-          return $status;
+          return $order;
         }
-        return $status;
+        // since $status < $network_confirmations payment_status should be 1 i.e. payment in progress if payment is not already completed
+        if ($order['payment_status'] != 2){
+            $order['payment_status'] = 1;
```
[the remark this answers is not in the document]
ankit61d: In the 2nd if statement, meta_key check is needed so that multiple callbacks don't update temp amount each time callback is hit. So, I have written it as
`if ($callback_status >= $network_confirmations && !metadata_exists('post',$wc_order->get_id(), $meta_key) ) {`
where `$meta_key = '_paid_to_'. $order['address'];`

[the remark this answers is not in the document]
DarrenWestwood: I do not believe this check is necessary since the payment status is also updated `$order['payment_status'] = 2;` on the 1st callback where the `$status >= $network_confirmations`. Consecutive callbacks should then return the existing order and will not update the paid amount again, due to the payment status check.
```
if ($order['payment_status'] == 2) {
    return $order;
}
```

php/Blockonomics.php
```diff
@@ -752,45 +820,69 @@ public function check_callback_secret($secret){
...
-    public function save_transaction($value, $order, $wc_order){
-        if (!metadata_exists('post',$wc_order->get_id(),'blockonomics_'. $order['crypto'] .'_txid', $order['txid']) )  {
-            update_post_meta($wc_order->get_id(), 'blockonomics_'. $order['crypto'] .'_txid', $order['txid']);
-            update_post_meta($wc_order->get_id(), 'expected_'. $order['crypto'], $order['satoshi']/1.0e8);
+    public function save_transaction($order, $wc_order){
+        $txid_meta_key = 'blockonomics_payments_txid';
+        $txid_meta_value = get_post_meta($order['order_id'], $txid_meta_key);
+        $txid = $order['txid'];
+        if (empty($txid_metavalue) || $txid_meta_value[0] == $txid){
+            update_post_meta($wc_order->get_id(), $txid_meta_key, $txid);
```
DarrenWestwood, line 828: We want the crypto to always be visible for all addresses, and not just for multiple payments:

1st Address


The same for transaction id's. The crypto for each transaction should always be visible. Transaction Id's are also currently overwriting, instead of adding to the metadata:

2nd Payment

The final custom fields should look as follows:

Full Display

shivaenigma: I believe problem with having 1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa (btc) is that 1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa is not searchable in the Woocommerce filters because it matches the complete string. To force it is match part of thing requires complex logic.

Ultimately we want it in the format you suggest @DarrenWestwood
_````1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa (btc), qr3fmxznghk8h0af7mpj0ay590ev5js72chef5md3w(bch)`_

But for now lets go with a simpler format of
_1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa, qr3fmxznghk8h0af7mpj0ay590ev5js72chef5md3w_

This way for single address/transaction payments (which is 95% cases), search will still work. For multiple address/transactions search won't work we document as an issue now and fix later

blockonomics-woocommerce.php
```diff
@@ -501,19 +508,30 @@ function blockonomics_activation_hook() {
...
-            maybe_drop_column($table_name, "timestamp", "ALTER TABLE $table_name DROP COLUMN timestamp");
-        }
+
+    // blockonomics_create_table() should only be run if there is no $installed_ver, refer https://github.com/blockonomics/woocommerce-plugin/issues/296
+    if (empty($installed_ver)){
         blockonomics_create_table();
+    } else if (version_compare( $installed_ver, $blockonomics_db_version, '!=')) {
+        blockonomics_run_db_updates($installed_ver);
     }
+        update_option( 'blockonomics_db_version', $blockonomics_db_version );
```
DarrenWestwood, line 524: `blockonomics_update_db_check` runs on every page. We should move updating the database version `update_option( 'blockonomics_db_version', $blockonomics_db_version );` to only run when required. After database creation, or after running updates

ankit61d:
This is how Custom Fields look like now:

image

blockonomics-woocommerce.php
```diff
@@ -464,24 +464,31 @@ function bnomics_async_scripts($url)
...
-        status int NOT NULL,
+        payment_status int NOT NULL,
         crypto varchar(3) NOT NULL,
         address varchar(191) NOT NULL,
-        satoshi int,
+        expected_satoshi int,
+        expected_fiat double,
         currency varchar(3),
-        value longtext,
+        paid_satoshi int,
```
DarrenWestwood, line 490: @shivaenigma, small note regarding the table structure: We should probably use BigInt for satoshi values instead of Int, although this will only affect really large payments.

DarrenWestwood (approved):
The code looks clean and tested everything is working well.

DarrenWestwood (approved):
Tested BigInt is working correctly to increase the max satoshi values that can be stored.

blockonomics-woocommerce.php
```diff
@@ -501,19 +508,30 @@ function blockonomics_activation_hook() {
...
+    global $blockonomics_db_version;
+    if (version_compare($installed_ver, '1.1', '<')){
         $table_name = $wpdb->prefix . 'blockonomics_orders';
-        if ($installed_ver < 1.1) {
-            maybe_drop_column($table_name, "time_remaining", "ALTER TABLE $table_name DROP COLUMN time_remaining");
-            maybe_drop_column($table_name, "timestamp", "ALTER TABLE $table_name DROP COLUMN timestamp");
-        }
+        maybe_drop_column($table_name, "time_remaining", "ALTER TABLE $table_name DROP COLUMN time_remaining");
+        maybe_drop_column($table_name, "timestamp", "ALTER TABLE $table_name DROP COLUMN timestamp");
+    }
```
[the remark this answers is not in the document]
DarrenWestwood: Retaining the upgrade scripts ensures that any updates to the plugin between any versions will function correctly and behave consistently. Retaining the upgrade from the previous version allows users to update from versions <= 3.5.7 to version 3.5.8, rather than the most recent version 3.5.9 if they so choose.

PR #318, opened 2023-04-06
Add partial payments checkbox in settings

ankit61d:
Based on feedback after release of version 3.6, this PR gives merchants an option to disable/enable the partial payments flow using coupons. By default, partial payments are enabled.

---

**Screenshots**:
Advance settings
partial_payments_checkbox

**Behavior if partial payments are disabled**: similar as was in prior to v3.6 i.e. status is set to Failed and Order note is added

partial_payment_disabled

Checkout when partial payments are disabled and order has been underpaid:
underpaid_checkout

Order note changed to show paid fiat amount in case of underpayment when partial payments are off:
image

---

Use cases tested:
1. Fresh Install ✔
2. Upgrade from v3.6 ✔
4. No JS mode ✔
5. Lite Mode ✔
6. No JS + Lite Mode ✔

### DIFF
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -400,6 +400,10 @@ class Blockonomics
         return get_option('blockonomics_lite', false);
     }
 
+    public function is_partial_payments_active(){
+        return get_option('blockonomics_partial_payments', false);
+    }
+
     public function is_error_template($template_name) {
         if (strpos($template_name, 'error') === 0) {
             return true;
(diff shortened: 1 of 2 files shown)


ankit61d:
Note: If partial payments are disabled, the `payment_status` is still 2, i.e. the db row for underpaid order looks like this
image
Here , order 235 is when partial payments are off and order 233 is when partial payments are enabled.

php/Blockonomics.php
```diff
@@ -400,6 +400,10 @@ public function is_lite_mode_active(){
         return get_option('blockonomics_lite', false);
     }
 
+    public function is_partial_payments_active(){
+        return get_option('blockonomics_partial_payments', false);
```
DarrenWestwood, line 404: Since we are setting the default to `true`, we may want to use the default as `true` here as well. This will only affect users who have not saved the Blockonomics settings, but have activated the plugin and using it.

DarrenWestwood (commented):
Functioning well apart from one small comment

Issue #292, opened 2022-10-11
JS syntax error when JS minify is enabled


ankit61d:
Based on customer issue
If the store has JS minify enabled, then checkout page shows rendering error due to syntax error in QR code generation

![image](https://user-images.githubusercontent.com/97018228/195025588-e5e06e77-ef3e-4d59-859a-87249080d7f3.png)
-----
#### Console error message: 
![image](https://user-images.githubusercontent.com/97018228/195025707-7476593c-0365-48be-8ab1-123038975162.png)
![image](https://user-images.githubusercontent.com/97018228/195025778-861a9b00-aafc-40ec-bc7f-2b15ddecd965.png)
-----
#### The syntax error line is 
<img width="734" alt="image" src="https://user-images.githubusercontent.com/97018228/195027039-c1a10194-fa50-44fe-8d32-d774b3799402.png">

PR #322, opened 2023-04-14
Removing Unused Variable, and Fixing Missing Semicolon and used Prettier to make it readable (#292)

aadyanthaya:
1) Used prettier to solve all the missing semicolons in the code.
2) Remove the const data variable.
3)Replaced the line value: data by value: this. data.payment_uri.

Fixes: #292  

CAUSE:
(Reason for the bug{js syntax error} is missing semicolon )

### DIFF
--- a/js/checkout.js
+++ b/js/checkout.js
@@ -1,98 +1,126 @@
 'use strict';
 class Blockonomics {
-
-    constructor({
-        checkout_id = 'blockonomics_checkout'
-    }={}) {
-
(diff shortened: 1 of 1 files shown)


DarrenWestwood (approved):
Further information: The issue lies with some minify scripts which do not handle the endlines after `${}` correctly during minification.
Using prettier automatically fixes these issues by correctly adding the `;` after these lines, and helps with code readability and consistency.

PR is working well

PR #325, opened 2023-04-25
Remove coupon order

aadyanthaya:
The coupon code logic  was removed and added paid as well as the remaining amount for #310
Order completion
image
Order completion email
image

PR #326, opened 2023-04-25
Disabled the partial payement

aadyanthaya:
With refernce to #325 #310 

Before disabling partial payment
image
After disabling partial payment
image


DarrenWestwood (approved):
Disables the partial payment setting by default.

PR #338, opened 2023-07-07
Blockonomics checkout page copy to clipboard

aadyanthaya:
https://github.com/blockonomics/woocommerce-plugin/assets/91294460/ba5dd90f-16e3-4611-bee3-7e70662f297f


DarrenWestwood:
We should load the copyToClipboard JS as a dependency for the checkout, as we are doing with qrious, and reconnecting-websocket:
https://github.com/blockonomics/woocommerce-plugin/blob/master/blockonomics-woocommerce.php#L520-L524
Let's get the existing copyToClipboard.js integrated, then we will work on the display fixes, such as icon colors and animations.

Glorydc650 (approved):

Issue #337, opened 2023-06-28
Payment method is confusing


blockonomics:
Even when paid by bitcoin cash, the payment method is always shown as *Bitcoin*

![image](https://github.com/blockonomics/woocommerce-plugin/assets/22165583/07496a80-755a-4e72-9a2a-c649e4cdbf40)
![image](https://github.com/blockonomics/woocommerce-plugin/assets/22165583/2be35a4b-e380-49fc-924a-429409020567)

PR #345, opened 2023-09-09
Updating payment method title based on the currency used in the order

aadyanthaya:
#337 (Bitcoin Cash)
image
(Bitcoin)
image

### DIFF
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -921,6 +921,7 @@ class Blockonomics
         $blockonomics_currencies = $this->getSupportedCurrencies();
         $selected_currency = $blockonomics_currencies[$order['crypto']];
         $wc_order->set_payment_method_title($selected_currency['name']);
+        $wc_order->save();
     }
 
     // Auto generate and apply coupon on underpaid callbacks
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -864,7 +864,6 @@ class Blockonomics
             if ($this->is_partial_payments_active()){
                 $this->add_note_on_underpayment($order, $wc_order);
                 $this->send_email_on_underpayment($order);
-                $wc_order->save;
             }
             else {
                 $wc_order->add_order_note(__(  wc_price($order['paid_fiat']). " paid via ".strtoupper($order['crypto'])." (Blockonomics).", 'blockonomics-bitcoin-payments' ));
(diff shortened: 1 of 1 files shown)


php/Blockonomics.php
```diff
@@ -917,6 +917,10 @@ public function process_callback($secret, $address, $status, $value, $txid, $rbf
         }
 
         $this->update_order($order);
+
+        $blockonomics_currencies = $this->getSupportedCurrencies();
+        $selected_currency = $blockonomics_currencies[$order['crypto']];
+        $wc_order->set_payment_method_title($selected_currency['name']);
```
aadyanthaya, line 923: Here we've found an edge case. Where if a customer does a partial payment with 2 different transactions of both BTC & BCH, then with the above logic, we'll be showing the payment method used in the last transaction.

To fix it, we'll need to make another query to fetch all the past transactions of the order and then compute the final title based on the different currency used.

Question is if it's worth doing an extra query for this edge case.

aadyanthaya:
```php
global $wpdb;

$query = $wpdb->prepare(
    "SELECT crypto FROM ".$wpdb->prefix."blockonomics_payments WHERE order_id = %d AND txid != ''",
    $order['order_id']
);

$results = $wpdb->get_results($query, ARRAY_A);

$blockonomics_currencies = $this->getSupportedCurrencies();
$selected_currency = $blockonomics_currencies[$order['crypto']];
$payment_method_title = $selected_currency['name'];

if ($results) {
    $cryptos = array_column($results, 'crypto');
    $cryptos[] = $order['crypto'];
    
    $cryptos_unique = array_unique($cryptos);

    if (in_array('btc', $cryptos_unique) && !in_array('bch', $cryptos_unique)) {
        $payment_method_title = $blockonomics_currencies['btc']['name'];
    } elseif (!in_array('btc', $cryptos_unique) && in_array('bch', $cryptos_unique)) {
        $payment_method_title = $blockonomics_currencies['bch']['name'];
    } elseif (in_array('btc', $cryptos_unique) && in_array('bch', $cryptos_unique)) {
        $payment_method_title = $blockonomics_currencies['btc']['name'] . ", " . $blockonomics_currencies['bch']['name'];
    }
}

$wc_order->set_payment_method_title($payment_method_title);
```

DarrenWestwood (approved):
Correctly sets the payment method title based on the most recent transaction.

PR #350, opened 2023-09-28
Resolves the bug  caused due to shortcode executing multiple times 

aadyanthaya:
Recently, we released WordPress 3.6.5 and encountered a bug where the shortcode is parsed or executed multiple times. This can occur for various reasons, such as the SEO plugin parsing the shortcode multiple times.
image

This fix ensures that the shortcode runs only once by making certain all files are included just a single time. The QR code for 'nojs' is executed only when 'no-js' is activated

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -84,18 +84,21 @@ function blockonomics_woocommerce_init()
     }
 
     function add_payment_page_shortcode() {
-        $show_order = isset($_GET["show_order"]) ? sanitize_text_field(wp_unslash($_GET['show_order'])) : "";
-        $crypto = isset($_GET["crypto"]) ? sanitize_key($_GET['crypto']) : "";
-        $select_crypto = isset($_GET["select_crypto"]) ? sanitize_text_field(wp_unslash($_GET['select_crypto'])) : "";
-        $blockonomics = new Blockonomics;
+        $currentFilter = current_filter();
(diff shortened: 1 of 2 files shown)


blockonomics-woocommerce.php
```diff
@@ -84,18 +84,21 @@ function bnomics_exclude_pages( $exclude ) {
     }
 
     function add_payment_page_shortcode() {
-        $show_order = isset($_GET["show_order"]) ? sanitize_text_field(wp_unslash($_GET['show_order'])) : "";
-        $crypto = isset($_GET["crypto"]) ? sanitize_key($_GET['crypto']) : "";
-        $select_crypto = isset($_GET["select_crypto"]) ? sanitize_text_field(wp_unslash($_GET['select_crypto'])) : "";
-        $blockonomics = new Blockonomics;
+        $currentFilter = current_filter();
+        if ($currentFilter == 'the_content') {
```
DarrenWestwood, line 88: The fix is working well at solving the issues with the All in One SEO plugin. Let's move this validation to the top of the shortcode function to improve readability
```
$currentFilter = current_filter();
if ($currentFilter != 'the_content') {
    return;
}
```

DarrenWestwood: Let's also add a code comment to explain why we are including this check.

DarrenWestwood (approved):

PR #352, opened 2023-10-19
Payment shortcode to be  not run on head

aadyanthaya:
This is to align with the new change where shortcodes are run multiple times 
So we do not want to run the shortcode if it is executed in the head section  to support all in SEO and Elementor executions


DarrenWestwood (approved):
Works well with both All in One SEO and Elementor

PR #353, opened 2023-10-28
Making plugin HPOS compatibility

aadyanthaya:


DarrenWestwood (commented):
Changes are looking neat and functioning as desired. Backward compatibility should not be a problem since the `update_meta_data` function was included since woocommerce v2.6.0.

DarrenWestwood (approved):
Changes are working well with different woocommerce versions and partial payments

PR #354, opened 2023-10-31
New settings page

ashthecoder05:
This PR is to move our current setting page from Settings > Blockonomics to Woocomerce > Settings > Payement > Bitcoin.

This is to take advantage of woocommerce and bring all our settings in one place.

Figma: https://www.figma.com/proto/mIJb7FXCn5JXjGLuSoeP1v/Blockonomics---Wordpress-Plugin---Onboarding?kind=&node-id=2-60&scaling=scale-down&starting-point-node-id=82%3A668&t=U5FVpebRrEOil4gQ-0&type=design

![image]image


shivaenigma:
https://github.com/blockonomics/woocommerce-plugin/assets/737092/395b8534-5291-4ed7-b1ce-18733b20fe1a

php/WC_Gateway_Blockonomics.php
```diff
@@ -122,17 +452,24 @@ public function handle_requests()
...
             $blockonomics->redirect_finish_order($order_id);
-        }else if ($get_amount && $crypto) {
+        } else if ($get_amount && $crypto) {
             $order_id = $blockonomics->decrypt_hash($get_amount);
             $blockonomics->get_order_amount_info($order_id, $crypto);
-        }else if ($secret && $addr && isset($status) && $value && $txid) {
+        } else if ($secret && $addr && isset($status) && $value && $txid) {
             $blockonomics->process_callback($secret, $addr, $status, $value, $txid, $rbf);
+        } else if ($test_setup) {
+            $blockonomics->settings_test_setup($api_key, $btc_active, $bch_active);
```
DarrenWestwood, line 472: It will be better to use the `wp_ajax_*` hook to handle the test setup request. This will ensure only logged in users are making the request and will keep this functionality separate from the payment processing logic:
`add_action( 'wp_ajax_blockonomics_test_setup', 'blockonomics_test_setup' );`
We will still be able to perform the request in vanilla js using `fetch()`.

DarrenWestwood:
Let's use a uniform method for displaying settings and their titles. The setting title should always be in bold and placed above the setting, as we have done for some settings. eg.
WooCommerce-settings-‹-Wordpress-Woocommerce-Blocks-—-WordPress (6)
Let's also ensure the inputs have the correct type, and the correct default value/placeholder text. Currently currency margin and slack % are using text type and api key and description do not have placeholder text.

DarrenWestwood (commented):
There are some issues with the temporary wallet UI. Please update as follows:
**No Payments Received**
No Funds Received

**Payments Received**
Received Funds

php/form_fields.php
```diff
@@ -0,0 +1,141 @@
...
+                'subtitle' => __('Enable Blockonomics plugin', 'blockonomics-bitcoin-payments'),
+                'type' => 'checkbox',
+                'label' => __('Enable Blockonomics as a payment method during checkout', 'blockonomics-bitcoin-payments'),
+                'default' => 'yes'
+            ),
+            'title' => array(
+                'subtitle' => __('Title', 'blockonomics-bitcoin-payments'),
+                'type' => 'text',
+                'description' => __('Payment method <i>title</i> displayed to the user during checkout.', 'blockonomics-bitcoin-payments'),
+                'default' => __('Blockonomics ', 'blockonomics-bitcoin-payments'),
```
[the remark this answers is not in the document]
DarrenWestwood: The default title is still set to `Blockonomics`: `'default' => __('Blockonomics ', 'blockonomics-bitcoin-payments'),`

php/WC_Gateway_Blockonomics.php
```diff
@@ -40,66 +42,389 @@ public function __construct()
...
+
+        $blockonomics = new Blockonomics;
+        $activeCurrencies = $blockonomics->getSupportedCurrencies();
+
+        foreach ($activeCurrencies as $code => $currency) {
+            $optionName = 'blockonomics_' . strtolower($code);
+            $isEnabled = $this->get_option($code . '_enabled') == 'yes' ? 1 : 0;
+            update_option($optionName, $isEnabled);
+        }
+        update_option('blockonomics_margin', (int)parent::get_option('extra_margin'));
```
DarrenWestwood, line 413: We can replace all cases of `parent::get_option` with `$this->get_option`

php/Blockonomics.php
```diff
@@ -361,11 +362,30 @@ public function testSetup()
         }
         return $test_results;
     }
+
+    public function settings_test_setup($api_key, $btc_enabled, $bch_enabled)
```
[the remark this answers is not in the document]
DarrenWestwood: Looks like we are still passing the settings values to test setup:
```
$api_key = isset($_GET["api_key"]) ? sanitize_text_field(wp_unslash($_GET['api_key'])) : "";
$btc_active = isset($_GET["btc_active"]) ? $_GET["btc_active"] : false;
$bch_active = isset($_GET["bch_active"]) ? $_GET["bch_active"] : false;
```
We want to instead fetch these from the saved settings, as we were doing before: https://github.com/blockonomics/woocommerce-plugin/blob/master/php/Blockonomics.php#L45-L58

`public function settings_test_setup($api_key, $btc_enabled, $bch_enabled)` should be `public function settings_test_setup()`

`$this->test_one_crypto($code, $api_key);` should be `$this->test_one_crypto($code);` etc.

We want to avoid passing the api_key, btc_active, bch_active as function params. These values should be collected from the saved settings values.

php/form_fields.php
```diff
@@ -0,0 +1,141 @@
...
+            'subtitle' => __('Partial Payments', 'blockonomics-bitcoin-payments'),
+            'label' => __('Allow customer to pay order via multiple payement  ', 'blockonomics-bitcoin-payments'),
+            'default' => $blockonomics->is_partial_payments_active() ? 'yes' : 'no',
+        );
+        $form_fields['network_confirmation'] = array(
+            'title' => __('', 'blockonomics-bitcoin-payments'),
+            'subtitle' => __('Network Confirmations', 'blockonomics-bitcoin-payments'),
+            'type' => 'select',
+            'description' => __('Network Confirmations required for payment to complete', 'blockonomics-bitcoin-payments'),
+            'default' => __(get_option('blockonomics_network_confirmation', 2), 'blockonomics-bitcoin-payments'),
```
DarrenWestwood, line 123: We are translating the saved value which is not required.
`'default' => __(get_option('blockonomics_network_confirmation', 2), 'blockonomics-bitcoin-payments')` should be `'default' => get_option('blockonomics_network_confirmation', 2)` as we have done for other settings.

DarrenWestwood: Same with the $callback_url. There is no need to use `__(..., 'blockonomics-bitcoin-payments')` to translate this value.

DarrenWestwood (commented):
Once a user has added an API Key, they should not be able to set the API Key to empty. This is to avoid the temp wallet being used again after a wallet has been setup.

DarrenWestwood:
On a fresh installation BTC should be enabled by default:
WooCommerce-settings-‹-Wordpress-Woocommerce-Blocks-—-WordPress (7)

We should also remove the saved settings from the woocommerce storage on uninstall `delete_option('woocommerce_blockonomics_settings')`

php/form_fields.php
```diff
@@ -0,0 +1,141 @@
...
+                'description' => __('Payment method for <i>bitcoin</i> displayed to the user during checkout.', 'blockonomics-bitcoin-payments'),
+                'default' => __('Bitcoin', 'blockonomics-bitcoin-payments'),
+                'placeholder' => __('Title', 'blockonomics-bitcoin-payments')
+            ),
+            'description' => array(
+                'subtitle' => __('Description', 'blockonomics-bitcoin-payments'),
+                'type' => 'text',
+                'description' => __('Payment method <i>description</i> displayed to the user during checkout.', 'blockonomics-bitcoin-payments'),
+                'default' => '',
+                'placeholder' => __('Desciption', 'blockonomics-bitcoin-payments')
```
DarrenWestwood, line 29: Description is spelt incorrectly.

DarrenWestwood (commented):
I have two minor concerns over how we are storing the settings:

1) Using two storage locations to save the settings is not a good practice (Wordpress and Woocommerce settings), however, I understand why we have implemented it this way to ensure all the old settings continue to work as desired without many changes.
In a separate PR based of this branch, we should investigate using only the woocommerce settings values throughout our code to avoid saving the Blockonomics settings in two places. The main concern is wether this change will work correctly for users updating from older versions, without the need to visit and save the new settings page first.

2) Since we are using form fields for the dividers and for sections which do not contain any settings, this leads to unnecessary field values being saved in the woocommerce settings.
All-the-WordPress-options-‹-Wordpress-6-0-—-WordPress
Can we remove the divider fields from the form array, and instead move dividers into the form fields using `generate_*_html`? eg. 
```
public function generate_text_html( $key, $data ) {
    $field_key = $this->get_field_key( $key );
    $defaults  = array(
        ...
        'add_divider' => false
    );

    $data = wp_parse_args( $data, $defaults );

    ob_start();
    ?>
    <tr valign="top">
        ...
    </tr>
    <?php
    if ($data['add_divider']==true) {
        ?>
            <tr valign="top">
                <td colspan="2"><hr /></td>
            </tr>
        <?php
    }
    return ob_get_clean();
}


$form_fields = array(
    ...
    'description' => array(
        'subtitle' => __('Description', 'blockonomics-bitcoin-payments'),
        'type' => 'text',
        'description' => __('Payment method <i>description</i> displayed to the user during checkout.', 'blockonomics-bitcoin-payments'),
        'default' => '',
        'placeholder' => __('Description', 'blockonomics-bitcoin-payments'),
        'add_divider' => true
    )
)
```
If either of these changes are going to add to much overhead, let me know in order to delay it for a later stage.

DarrenWestwood (commented):
The link in activation notification is incorrect
Plugins-‹-Wordpress-6-0-—-WordPress

anktd:
Tested this version. New settings page is looking good 👍 
There are few things to be looked at:

- Getting php warning msg when activating the plugin (logs below image)
image

```
[30-May-2024 08:03:51 UTC] PHP Warning:  Creating default object from empty value in /Applications/MAMP/htdocs/wpmore/wp-content/plugins/woocommerce-plugin-new-settings-page/php/Blockonomics.php on line 241
[30-May-2024 08:03:51 UTC] PHP Notice:  Undefined property: stdClass::$message in /Applications/MAMP/htdocs/wpmore/wp-content/plugins/woocommerce-plugin-new-settings-page/blockonomics-woocommerce.php on line 239
[30-May-2024 08:04:46 UTC] PHP Warning:  include_once(Blockonomics.php): failed to open stream: No such file or directory in /Applications/MAMP/htdocs/wpmore/wp-content/plugins/woocommerce-plugin-new-settings-page/blockonomics-woocommerce.php on line 100
[30-May-2024 08:04:46 UTC] PHP Warning:  include_once(): Failed opening 'Blockonomics.php' for inclusion (include_path='.:/Applications/MAMP/bin/php/php7.4.33/lib/php') in /Applications/MAMP/htdocs/wpmore/wp-content/plugins/woocommerce-plugin-new-settings-page/blockonomics-woocommerce.php on line 100
[30-May-2024 08:05:00 UTC] PHP Warning:  include_once(Blockonomics.php): failed to open stream: No such file or directory in /Applications/MAMP/htdocs/wpmore/wp-content/plugins/woocommerce-plugin-new-settings-page/blockonomics-woocommerce.php on line 100
[30-May-2024 08:05:00 UTC] PHP Warning:  include_once(): Failed opening 'Blockonomics.php' for inclusion (include_path='.:/Applications/MAMP/bin/php/php7.4.33/lib/php') in /Applications/MAMP/htdocs/wpmore/wp-content/plugins/woocommerce-plugin-new-settings-page/blockonomics-woocommerce.php on line 100
```

- Test setup for BCH is showing success even though callback url is not same on plugin as on the store, which results in error on checkout page
image

image

- Description for Store row reads "To use your own wallet and start withdrawing fund,you can setup a Blockonomics store". Lets update this now that temp wallets are removed. It could something like "Setup a Blockonomics store to start accepting payments in your wallet"
image


We should also sync this with master branch as this branch is behind on the recent commits.

anktd:
Changes with test setup are tested and working fine. 

1. Empty API Key should not be saved since temp wallets are removed. Currently it is allowing user to save empty API key after fresh installation and displays the msg 'Your settings have been saved'.
 Once an API Key is added & saved, it is working as intended and not saving the empty API key again. 
image

2. Checkout with No Javascript setting activated doesn't have centrally aligned QR Code and 'Click here if already paid' button. We can address this later if its not straightforward solution.
image


3. Uninstallation of plugin is giving a Php warning `PHP Notice:  Function wpdb::prepare was called incorrectly. The query argument of wpdb::prepare() must have a placeholder. `

css/admin.css
```diff
@@ -0,0 +1,92 @@
...
+#temp-wallet-notification-box {
+    border: 1px solid #8FBC8F;
+    background-color: #F0FFF0;
+    color: #006400;
+    padding: 10px;
+    margin: 10px 0;
+    border-radius: 4px;
+    display: flex;
+    align-items: center;
+}
```
anktd, line 48: We can remove this block since it is used nowhere.

css/admin.css
```diff
@@ -0,0 +1,92 @@
...
+    margin: 10px 0;
+    border-radius: 4px;
+    display: flex;
+    align-items: center;
+    display: none;
+}
+
+#temp-wallet-input {
+    cursor: pointer;
+}
```
anktd, line 64: This can be removed too.

anktd (changes requested):
Setting page is looking good. There are a couple of unused blocks we can remove.

anktd (approved):

shivaenigma:
Looks good to one. One thing to fix is that we cannot enabled until there is an [APIKey set](https://nimb.ws/QMlon6N).  Otherwise user is seeing the option and unecessary error of unable to generate address

php/form_fields.php
```diff
@@ -0,0 +1,118 @@
...
+            'type' => 'checkbox',
+            'subtitle' => __('No Javascript checkout page', 'blockonomics-bitcoin-payments'),
+            'label' => __('Enable this if you have majority customer that uses tor like browser that blocks JS', 'blockonomics-bitcoin-payments'),
+            'default' => get_option('blockonomics_nojs') == 1 ? 'yes' : 'no',
+        );
+        $form_fields['partial_payment'] = array(
+            'title' => __('', 'blockonomics-bitcoin-payments'),
+            'type' => 'checkbox',
+            'subtitle' => __('Partial Payments', 'blockonomics-bitcoin-payments'),
+            'label' => __('Allow customer to pay order via multiple payement  ', 'blockonomics-bitcoin-payments'),
```
DarrenWestwood, line 92: Spelling issue of 'payment'

php/form_fields.php
```diff
@@ -0,0 +1,118 @@
...
+                '2' => __('2(Recommended)', 'blockonomics-bitcoin-payments'),
+                '1' => __('1', 'blockonomics-bitcoin-payments'),
+                '0' => __('0', 'blockonomics-bitcoin-payments'),
+            ),
+        );
+        $form_fields['call_backurls'] = array(
+            'title' => __('', 'blockonomics-bitcoin-payments'),
+            'type' => 'text',
+            'description' => __('You need this callback URL to setup multiple stores', 'blockonomics-bitcoin-payments'),
+            'subtitle' => __('Callback URL :', 'blockonomics-bitcoin-payments'),
```
DarrenWestwood, line 111: No need for ':'. All other titles do not have colons.

DarrenWestwood (commented):
Code looks good apart from 2 minor issues noted.

shivaenigma:
Error on test setup page is wrong ```Navigate to Settings > Blockonomics > Currencies and click Test Setup`

DarrenWestwood (approved):
New logic to enable the plugin when the API Key changes is working well.

DarrenWestwood (approved):
Updated logic during upgrade, for existing users, is working as desired

PR #356, opened 2023-11-01
Fixing additional script bug on block theme

aadyanthaya:
## What happened?
- With our new change for payment page as shortcode, additional scripts which incldues `blockonomics_data` is not added for block based themes like Twentytwentythree.

## Impact?
All block based themes cannot load the checkout page as it shows a Display Error and has a JS error in the browser console.

## Fix?
To fix, instead of using `wp_add_inline_script` to add the script we need to use `wp_footer` action by adding the additional script at the bottom of the page.


DarrenWestwood (approved):
Changes are clean and have been tested with multiple themes by Ark and I.

PR #358, opened 2023-11-28
Fixed JS script url and added defer attribute

aadyanthaya:
Previously we used `clear_url` filter hook to added the defer attribute to the script tag. Which recently has been deprecated and is generating a invalid url for scripts. It looks something like this

```html
<script src="js/checkout.js?ver=3.6.7' defer='defer'"></script>
```

Now changed this by passing a stratergy `defer` to the `wp_register_script()`  and this creates the url correctly as below

```html
<script src="js/checkout.js?ver=3.6.9" defer="defer" data-wp-strategy="defer"></script>
```


DarrenWestwood (approved):
The changes are working well with latest versions of Wordpress and Woocommerce.
Since the [strategy parameter](https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/) was introduced in Wordpress 6.3, scripts will only be deferred with Wordpress versions greater than 6.3. The plugin will still work correctly with older version, but scripts will not be deferred, but still loaded in the footer.

Issue #361, opened 2023-12-27
Wrong value displayed for bitcoin currency type  in woocommerce 


ashthecoder05:
Problem: When the currency type is bitcoin in the order paid page at the transaction details level after the payment is made it is showing 0.00000000 instead of the actual amount paid. As you can see in the image below.

PR #362, opened 2023-12-28
bitcoin currency decimal issue

aadyanthaya:
The proper value was not displayed at the payment details level

image

Solution  for the issue #361


DarrenWestwood (approved):
Clean fix and works well. Decimals correctly formatted according to store decimals setting.

PR #364, opened 2024-01-29
Connect_to_Wallet

ashthecoder05:
created a button to connect to any wallet add-ons


js/checkout.js
```diff
@@ -41,6 +41,20 @@ class Blockonomics {
...
         this._display_error_wrapper.style.display = 'none';
+
+        this.wallet();
+    }
+
+    async wallet() {
+        const connectBtn = document.getElementById('connect-wallet');
+
+        connectBtn.addEventListener('click', async () => {
+            const provider = new ethers.providers.Web3Provider(window.ethereum);
```
DarrenWestwood, line 52: Using the injected `window.ethereum` will work for browser extensions.

To support web app wallets (Coinbase), hardware wallets (Ledger), and mobile wallets (Trust, Metamask mobile app), we can consider using an existing package like [Web3Modal](https://walletconnect.com/web3modal) ([Demo](https://web3modal.com/)). This will support a larger number wallets and avoids us dealing with connectors for each wallet type.

PR #376, opened 2024-07-15
Fix headers already sent error in version 3.7.4

anktd:
The debug log shared by merchant [here](https://wordpress.org/support/topic/latest-update-prevented-admin-login-account-creation/) shows a php warning about "Headers already sent"

`PHP Warning: Cannot modify header information - headers already sent by (output started at /homepages/44/MDSP/htdocs/wordpress/wp-content/plugins/blockonomics-bitcoin-payments/php/form_fields.php:1) in /homepages/44/MDSP/htdocs/wordpress/wp-includes/http.php on line 505
`

This is caused by [empty line before php tag](https://github.com/blockonomics/woocommerce-plugin/blob/4e9fa04a622afadc7767286ad2807598338dcd4f/php/form_fields.php#L1) declaration in `php/form_fields.php` file. This is causing php functions that need to send HTTP headers (like setcookie(), session_start(), or header()) to fail because content has already been sent to the browser.

This file also had missing ABSPATH check. Without this check, if someone knows the file's URL, they could potentially access and execute the file directly, making this a potential security risk. I've added this statement in this PR to fix this.

### DIFF
--- a/php/Blockonomics.php
+++ b/php/Blockonomics.php
@@ -38,8 +38,9 @@ class Blockonomics
         return $rounded_total_paid_fiats;
 
     }
-    
-    
+
+    private $api_key;
+
     public function __construct()
     {
         $this->api_key = $this->get_api_key();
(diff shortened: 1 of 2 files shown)


anktd:
Added fix from https://github.com/blockonomics/woocommerce-plugin/pull/375 about the deprecation warning for `api_key` variable for websites using php version 8.2 onwards.

PR #378, opened 2024-10-14
Move BCH to advanced settings

anktd:
This PR is part 1 of supporting the new user flow in plugin, and making it in line with new Blockonomics Rest API. 
Stuff done:
- Currencies Tab is removed
- Enable BCH is now a checkbox under Advanced Settings (default behavior is unchecked)
- Test Setup doesn't support BCH now
- BCH checkout is still supported if BCH is enabled under Advanced Settings
- Revised unit tests to align with the removal of BCH from the test setup process


anktd:
Closing this as #379 contains changes from this PR as well.

PR #379, opened 2024-10-24
Implement v2 API endpoints

anktd:
- This PR is one of the part to support the new user flow in plugin, and making it in line with v2 API.
- Other part is #380 
- Comparision for changes in both these PR → [here](https://github.com/ashthecoder05/woocommerce-plugin/compare/setup-page...anktd:woocommerce-plugin:new-api-endpoints)
- Design doc on notion for new final design → [here](https://www.notion.so/blockonomics/Woocommerce-Plugin-APIv2-User-Flow-Study-92d8039a6c4846a89964a0e10552d115#65efd66c7ac04a808cf8a574097fe7d4)
- Testing sheet on notion → [here](https://www.notion.so/blockonomics/54bad8b035134f668d63c9f5114c4819?v=b545daa0eedf4be887b9a59caab264af)

---

**Changes done in this PR:**
- Currencies tab is removed from settings page
- Enable BCH is now a checkbox under Advanced Settings (default behavior is unchecked)
- Test Setup doesn't support BCH now
- BCH checkout is still supported if BCH is enabled under Advanced Settings
- Save API button is removed. Test Setup saves API key and then runs test setup
- Store Name and Enabled Cryptos are saved as metadata in WP when test setup is success and displayed above API Key input box
image

- `new_address` generation now uses `match_callback` parameter.
- refactored `js/admin.js` to improve readability

---
**System side flow for Test Setup**:

image

---
**Challenges / Concerns with this new flow**:
1. Since Test setup and save API key is only one button, behavior for when both API Key and any other setting is changed is unclear. Currently it saves everything in this scenario. When only other settings have changed and not API Key, this button will not work and it asks to use 'Save Changes' button.
Kapture 2024-10-30 at 07 55 51

2. Metadata displayed for Store name and Enabled cryptos is only shown after page is refreshed once Test Setup is hit.
Kapture 2024-10-30 at 07 59 42

3. 'Enabled Crypto' metadata display doesn't support BCH even if it is checked under Advanced Settings.
4. Unittests for Test setup are broken after v2 endpoints are used. I will create follow up PR for that.


shivaenigma:
>Challenges / Concerns with this new flow

Understood. All are minor concerns and I don't see them stopping the release

shivaenigma:
**Guidance for test setup**
For now test setup flow looks good. For future test setup,  have minimal code/automation to check stores/set callback. User is expected to himself copy/paste callback URL in stores UI. Once we [improve server side responses](https://github.com/Blockonomics-Dev-Team/Blockonomics/issues/3230), we can remove the extra code in test setup

**Guidance for wizard**
Wizard is expected to do the automations like creating store/ associating wallet with store etc

PR #395, opened 2025-06-09
Add minor copy changes to setup wizard

cnohall:
Some simple copy changes to the setup wizard to make it clearer

### DIFF
--- a/php/admin-page.php
+++ b/php/admin-page.php
@@ -106,8 +106,8 @@ function blockonomics_setup_page() {
                 </div>
                 <ol>
                     <li><a href="https://www.blockonomics.co/register" target="_blank">Sign up</a> to Blockonomics</li>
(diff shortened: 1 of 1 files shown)


php/admin-page.php
```diff
@@ -177,7 +177,7 @@ function blockonomics_setup_page() {
                 <?php else: ?>
                     <!-- Final Success Screen -->
                     <div class="bnomics-wizard-heading">
-                        <h3>Congrats! Your store <?php echo esc_html(strtoupper($store_name)); ?> setup is complete!</h3>
+                        <h3>Congrats! Your store <?php echo esc_html(strtoupper($store_name)); ?> setup is ready!</h3>
```
anktd, line 180: should we also remove `strtoupper` from this? It do stick out somewhat 🫤

cnohall: Good call, I updated the copy slightly as well. 

image

Please let me know if you have any tips of further improvements

PR #396, opened 2025-06-17
[DRAFT] [Feature] Add Bitcoin Discount Percentage option to advanced settings

cnohall:
image


cnohall:
May I ask for your review of this @anktd? I've done some testing, but haven't used the plugin in a long time. Please let me know if you think of something that I missed when testing.

PR #400, opened 2025-09-01
Include USDT payments

DarrenWestwood:
Includes USDT payments. 
Done:
- All checkout, payment, and callback functionality

Still todo:
- Update the test setup/checkout to check the store setup and enable/disable USDT accordingly
- Refactor multiplier functionality to a single location in the supported crypto data:
```
'usdt' => array(
        'code' => 'usdt',
        'name' => 'USDT',
        'decimals' => 6
)
```


php/Blockonomics.php
```diff
@@ -232,197 +277,254 @@ private function set_headers($api_key)
...
+            update_option("blockonomics_store_name", $store_name);
+        }
+    }
+
+    private function test_cryptos($enabled_cryptos) {
+        $success_messages = [];
+        $error_messages = [];
+
+        foreach ($enabled_cryptos as $code) {
+            $response = $this->new_address($code, true);
```
anktd: We are using the reset=1 param when doing test setup, this has problems with BTC:
1. Primary key error on subsequent checkout
2. Test setup resulting in false positives (say, instead of displaying gap limit error)

DarrenWestwood:
1. Updated usdt txn link on Order Confirmation page
2. Updated handling of error messages. This will also be required in the WHMCS module: https://github.com/blockonomics/whmcs-bitcoin-plugin/blob/master/modules/gateways/blockonomics/blockonomics.php#L465-L472
3. Removed auto update callback url if partial match

PR #421, opened 2026-05-15
Add floating Save Changes button to Blockonomics settings page

TandonSuyash:
The Save Changes button on the Blockonomics settings page was only 
visible at the bottom, requiring the user to scroll down to save. 

This PR adds a floating Save Changes button that:
- Stays fixed on screen while scrolling
- Is disabled (grey) by default, enabled only when a field is modified
- Hides the original button only after JS confirms everything is working,
  keeping it as a silent fallback

### DIFF
--- a/blockonomics-woocommerce.php
+++ b/blockonomics-woocommerce.php
@@ -109,7 +109,6 @@ function blockonomics_woocommerce_init()
     }
 
     add_action( 'admin_enqueue_scripts', 'blockonomics_enqueue_custom_admin_style' );
-    add_action( 'admin_footer', 'blockonomics_floating_save_btn' );
     add_action( 'wp_ajax_test_setup', 'blockonomics_test_setup' );
 
     function bnomics_exclude_pages( $exclude ) {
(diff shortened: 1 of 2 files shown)


blockonomics-woocommerce.php
anktd, line 129: This page already enqueues `css/admin.css` and `js/admin.js`. Why did we add inline CSS/JS instead of using these?

blockonomics-woocommerce.php
```diff
@@ -143,6 +144,106 @@ function blockonomics_enqueue_custom_admin_style() {
...
+            @media (max-width: 960px) {
+                #blockonomics-floating-save {
+                    left: 10px;
+                }
+            }
+        </style>
+
+        <button type="button" id="blockonomics-floating-save" disabled>
+            Save changes
+        </button>
```
anktd: Why are we adding a second button instead of styling the existing `.woocommerce-save-button` directly What's the minimum CSS needed to make the existing button float? Is JS needed at all if we just make it sticky? Thinking along these lines will help with a solution.

blockonomics-woocommerce.php
```diff
@@ -143,6 +144,106 @@ function blockonomics_enqueue_custom_admin_style() {
...
+            #blockonomics-floating-save:disabled {
+                background: #a0a0a0;
+                cursor: not-allowed;
+                box-shadow: none;
+            }
+
+            /* Hide the default save button after the floating button is initialized. */
+            body.blockonomics-floating-save-ready .woocommerce-save-button {
+                display: none !important;
+            }
```
anktd: This is very hacky code. We're rendering a new button, then hiding the real one with !`important`, then forwarding clicks back to the real one via JS.

blockonomics-woocommerce.php
```diff
@@ -143,6 +144,106 @@ function blockonomics_enqueue_custom_admin_style() {
...
+               // Use WooCommerce's existing save action.
+                floatingButton.addEventListener('click', function () {
+                    if (floatingButton.disabled) {
+                        return; 
+                    }
+                    saveButton.click();
+                });
+
+            }());
+        </script>
```
anktd: Do we need any of this JS? What things this is solving we cant do from css alone.

blockonomics-woocommerce.php
```diff
@@ -143,6 +144,106 @@ function blockonomics_enqueue_custom_admin_style() {
...
 
+    function blockonomics_floating_save_btn() {
+
+        $page    = isset( $_GET['page'] )    ? sanitize_text_field( wp_unslash( $_GET['page'] ) )    : '';
+        $tab     = isset( $_GET['tab'] )     ? sanitize_text_field( wp_unslash( $_GET['tab'] ) )     : '';
+        $section = isset( $_GET['section'] ) ? sanitize_text_field( wp_unslash( $_GET['section'] ) ) : '';
+
+        if ( 'wc-settings' !== $page || 'checkout' !== $tab || 'blockonomics' !== $section ) {
+            return;
+        }
```
anktd: Duplicate code. These checks already exist above.

anktd:
Before another round of changes, I'd like you to step back and rethink the approach from scratch. Smallest change that solves the issue is our goal for the PR.  Aim of this task is for you to learn plugin structure, and not to just ship faster. 
Please redo this from scratch with the existing codebase in mind.

anktd:
image
🐛  -> Save button is getting parked on top of full-width admin menu items

TandonSuyash:
image
image
I tried reproducing the bug twice, but it is placed as expected for my setup.
Can you please share how you were able to reproduce the issue ?

css/admin.css
```diff
@@ -82,3 +82,18 @@
...
+/* Floating save button */
+.woocommerce-save-button {
+    position: fixed;
+    bottom: 24px;
+    left: 200px;
+    z-index: 9999;
+    box-shadow: 0 2px 8px rgba(0, 0, 0, 0.25);
+}
+
+@media (max-width: 960px) {
```
anktd: 782px is WP's standard mobile breakpoint. Change value from 960px to 782px

anktd (changes requested):

